PR #11, opened 2013-09-12
Adding support for non nested geometries.

eolivi:
retrocompatibility assured for geom files (not for previous matrices).


src/presmooth.cpp
```diff
@@ -37,18 +37,18 @@ The fact that you are presently reading this means that you have had
...
 */
 
-#include "symmatrix.h"
-#include "vector.h"
-#include "mesh3.h"
-#include "sparse_matrix.h"
-#include "fast_sparse_matrix.h"
+#include <symmatrix.h>
+#include <vector.h>
+#include <mesh.h>
```
massich: This broke `presmooth.cpp`. 

`SparseMatrix` in `mesh3.h` had `.gradient(void)` method while there is no similar signature in `mesh.h` braking ` SparseMatrix SmoothMatrix = SourceMesh.gradient();`

PR #63, opened 2014-08-19
Multiple nonconductive

papadop:
First to gather comments.


papadop:
Where are we on this front ? Can we merge ?

dang-kai:
The code and tests are finished. But the analytical solution to compare with is not ready yet. I think we can merge once the analytical solution is added.

agramfort:
what's up with this branch?

dang-kai:
I think we can remove this branch since all the commits have been moved to the "handle zero conductivity" branch and merged.

papadop:
Closed....

PR #72, opened 2015-04-05
Enable NeuroDebian and its -devel for libmatio library

yarikoptic:


agramfort:
thanks @yarikoptic 

I think you should rebase on master and force push on this PR.

we'd get latest cmake.

thanks heaps for helping on this.

PR #73, opened 2015-04-23
Update README.rst

lazyparser:
The users using RedHat/CentOS or related Linux distributions
may encounter a 'shared libraries not found' issue.
This PR provides an one-line solutioln for this issue.


papadop:
Sorry I only see this pull request now. Incorporated. Thank you very much.

PR #84, opened 2016-03-07
Fused super project

papadop:
Basically I fused the Superproject into OpenMEEG proper. This will allow to modify them in sync (in the same branch). To do that, I did (from master) (well not exactly that, but it should be equivalent to) :
git checkout -b FusedSuperProject   
mkdir OpenMEEG
git add OpenMEEG
git mv AUTHORS ChangeLog ci cmake CMakeLists.txt coding_guidelines.txt Copyright.txt CTestConfig.cmake data documentation examples file_header.txt KWStyleFilesToCheck.txt libs LICENSE.txt NEWS OpenMEEGConfig.kws.xml OpenMEEGConfigure.h.in packaging README.rst tests tools win32addons Wrapping .travis.yml OpenMEEG
git commit -m "Moved files to OpenMEEG subdir." -a
git remote -f add superproject git@github.com:openmeeg/super-project.git
git merge superproject/master
A simple edition (cmake/externals/projects_modules/)

Of course, once this is pulled the super-project.git has to be closed/removed or at leat for now emptied with a README referring to this repo....


agramfort:
@papadop git history is a mess. There are 905 files changed and the diff is not readable on github. How can I review this?

papadop:
I described almost exactly what I did. The only thing that was not in the super-project is the last commit "Modify the external project so that it does not download OpenMEEG any…". If you now a better way to do the merge without loosing the history, tell me. Otherwise, I do not see any other way.... Functionally, all other commits were already there in the super-project, so no need to review them. The main thing is: do we agree that such a merge is useful or not. I believe it isor otherwise foreach branch of OpenMEEG, we eventually need to have a branch for the superproject, which is a pain. There might be some further  cleanups that can be done (sharing the version number, remove somw redundant stuff in cmake subdirectories, ...) but this can wait. After all there is still some code in camek to extract the version number from svn !!! (seen that the other day, need to remove/adapt, knowing that give does not have global rev numbers).

agramfort:
I am relunctant to do this as it will further increase the difficulty of
packaging this code on linux distributions. Besides all the super project
is motivated by the matio support (hdf5, etc...)
which is fragile and I feel is an overkill for storing a single matrix per
file. It's a lot
of complexity and maintainance burden for almost no practical benefit. If
you're asking
what I suggest, well the Numpy file format (.npy) has become a standard and
it's really
simple and well documented.

http://docs.scipy.org/doc/numpy-dev/neps/npy-format.html

people have wrote matlab code to read such files:

https://github.com/kwikteam/npy-matlab

there is a super light c++ lib with CMake support too:

https://github.com/rogersce/cnpy

my 2c

papadop:
Alexandre Gramfort wrote:
> I am reluctant to do this as it will further increase the difficulty of
> packaging this code on linux distributions.

No this is not true. OpenMEEG proper remains a subproject which can be build separately. It is only the git repo that are merged. But I agree that the documentation must be clear that there are two ways to build OpenMEEG (one from the superproject, and another one at the OpenMEGG level. I must admit that the layout I proposed might have the problem that people not reading any README will start from the toplevel so with the superproject.... This can maybe be corrected, but actually I do not know exactly what I want in this respect...

> Besides all the super project
> is motivated by the matio support (hdf5, etc...)
> which is fragile and I feel is an overkill for storing a single matrix per
> file. 

Not exactly.... The superproject came from the idea (and you proposed it) to help users to built OpenMEEG even when they do not have the possibility of installing things on their machines. The main reason of the superproject is lapack not mation (and actually we can still build without matio with the TrivialBinIO). So the need of the superproject remains even if we decided to ditch matio.

> It's a lot
> of complexity and maintainance burden for almost no practical benefit. If
> you're asking
> what I suggest, well the Numpy file format (.npy) has become a standard and
> it's really
> simple and well documented.

Well it has saved us in a few circumstances (bigger than 4Gb files, numerous bugs with TrivialBinIO). And provides better timings with big files (loading huge files takes more time than decompressing them).

The numpy format could have been a nice alternative, but as far as I can tell there is no symmetric or sparse matrix support (at least not in cnpy). Moreover, it loads the data as char* which will also will create some problems (dealing with endianness, dealing with files generated by numpy in some weird cases (float32 or sparse matrices which are handled by dumping a pickle file as far as I understood).

I'm not against adding support for the numpy format, I even can help. But for the time being, this will be along matio not in replacement of matio. At the very least until some point where we are confident enough about the implementation. But even in the case we totally remove matio, the superproject will be necessary for lapack, for cnpy and zlib at least (and I'm inclined to add another library to this list soon, but more of that later after doing some tests).

So really, the purpose of the superproject is totally un-related to matio, so let's do not mix things....

That being said, I'll retire the current merge request as it will make the merging of Kai's work (or any other people with sizeable branches, but I do not think there are other and all the usual suspects will probably read this message) more complicated.

So my intent is first to merge Kai's work and then proceed with the merge of the superproject and OpenMEEG proper. I'm open to discussions on the tree layout after the merge (we can put the superproject in a subdir of OpenMEEG or keep the hierarchy I used or use sthg else) or in ways of doing it better than I did in the proposal.

PR #85, opened 2016-03-10
C++11 and numpy 1.7 API

papadop:
2 commits in a same pull request.
The first make us require a C++-11 compiler. This homogeneize the code between unix/Mac and Visual C++ (which is by default since VS 12). C++-11 support is available (but with a flag) in clang and g++ since version 4.8. Note that we use nothing particular to C++-11, but there are some library incompatibilities in the conversions of file streams and results of read on streams to bool (now the cast has to be made explicitely).
The second change removes some warnings coming from openmeeg python support. I activated the numpy strict API which is available since numpy 1.7 (currently at 1.10) and modified the code appropriately (mainly use some macros to access fields of the numpy structure + some small name changes).

### DIFF
--- a/Wrapping/src/numpy.i
+++ b/Wrapping/src/numpy.i
@@ -7,319 +7,573 @@
 #endif
 #include "stdio.h"
 #include <numpy/arrayobject.h>
+%}
+
+/**********************************************************************/
+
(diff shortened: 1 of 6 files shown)


agramfort:
LGTM merging. thanks

PR #86, opened 2016-03-10
Zero cond branch

papadop:
This is a merge of Kai zero conductivity handling.... Regression tests have been added (correctness tests are still needed ;-) ).

I do not know why commits by other people than Kai and Maureen appear in the commits, but those are already on the master and since github claims that it can do the merge automatically, I guess this is OK.


agramfort:
see #88

I basically cleanup the history so we have a single clean commit on top of master.

I'll let you merge if you're happy. Then close this PR

PR #89, opened 2016-03-11
Bug fix for sensors with a radius

dang-kai:
The nearest triangle is put into the list twice when searching triangles for sensors with a radius, which makes the total electrode area incorrect.


agramfort:
LGTM

PR #108, opened 2016-05-07
rm python viz code

### LINKED ISSUE #102: installs om_display.py (outside of openmeeg python pkg) but I don't even see it used
```python
$> git grep om_display
OpenMEEG/tools/python/CMakeLists.txt:    install(FILES om_display.py
OpenMEEG/tools/python/om_display.py:this provides two functions om_display_vtp, and om_display_vtk.
OpenMEEG/tools/python/om_display.py:from om_display import om_display_vtp # visualiation with VTK
OpenMEEG/tools/python/om_display.py:def om_display_vtp(f, n = 0):
OpenMEEG/tools/python/om_display.py:def om_display_vtk(f,d = 0,n = 0):
```

agramfort:
moved python viz code to its own repo:

https://github.com/openmeeg/openmeeg_viz

closes #102


agramfort:
@bburle is that fine with you?

I want to move all the python viz code out of openmeeg so we can have a real python package evently on pypi.

papadop:
Not sure I like the move.... Again two repos that we will have to keep in sync. No  possibility to have this within the OpenMEEG repo  and still have a proper pypi packaging ??

agramfort:
>
> Not sure I like the move.... Again two repos that we will have to keep in
> sync. No possibility to have this within the OpenMEEG repo and still have a
> proper pypi packaging ??
>
putting openmeeg on pypi  means rewriting the build system to use setup.py
files etc.
this file is not even a python package nor a module in the openmeeg
namespace.
it's slowing updates for little gain. See debian issue.

I make hard decisions here because we must release. We have not done it for
ages and I need to focus on bug fixes (they are still a few...). We'll add
features when we need them. This time we'll make sure we have tests run on
travis etc. :)

PR #109, opened 2016-05-10
Correct copy of array.

papadop:
Correct the python regression.


agramfort:
thx @papadop 

is it possible to add a test as python code for this?

papadop:
Not sure to understand. If you mean the bug correction, it is already covered by the python test. If it is the other bug mentioned with temps, I can easily make a test, but for the time being it will fail.

agramfort:
I thought what you wrote with the getcol copy/ref count was a simple unit test but maybe it's already there. If so +1 for merge.

thx @papadop

PR #111, opened 2016-05-15
FIX clang warnings

agramfort:


OpenMEEG/libs/OpenMEEG/src/assembleSourceMat.cpp
```diff
@@ -118,27 +118,26 @@ namespace OpenMEEG {
                 domain = geo.domain(domain_name);
             }
             const double sigma = domain.sigma();
-	    //only consider dipoles in non-zero conductivity domain
-	    if(!sigma==0)
```
agramfort: clang was complaining about this

OpenMEEG/libs/OpenMEEG/src/geometry.cpp
```diff
@@ -186,9 +186,10 @@ namespace OpenMEEG {
                         nb_current_barrier_triangles()+=mit->nb_triangles();
                         for (Mesh::iterator tit=mit->begin(); tit!=mit->end(); ++tit)
                             tit->index() = index++;
-                    }else
+                    } else {
```
agramfort: clang was complaining about potential dangling else

agramfort:
@papadop merge if you're happy

OpenMEEG/libs/OpenMEEG/src/assembleSourceMat.cpp
papadop: This cannot be the proper fix.... if (sigma!=0.0) ???
I do not know the clang warning, but this seems rather valid code (even if not the most readable). I Can see two possible complaints : The ! == or the fact that 0 is an integer constant and not a double one. The above proposal fixes both.... I have no access to your branch, otherwise I would have done the change myself.

agramfort: sorry what do you suggest as fix?

papadop: if (sigma!=0.0) ... Ah, OK ,  the hunk was splitted at the wrong place, this is in essence what you did....

PR #112, opened 2016-05-15
add travis config to use system libs

agramfort:


agramfort:
@papadop link pb with system libs reproduced.

See https://travis-ci.org/openmeeg/openmeeg/jobs/130360319

@yarikoptic any idea?

OpenMEEG/CMakeLists.txt
```diff
@@ -185,7 +185,8 @@ add_subdirectory(Wrapping)
 #   Tests
 
 if (BUILD_TESTING)
-    include(CreateTestEnvironment)
+    # XXX : this breaks when trying to build OpenMEEG without the project :(
+    # include(CreateTestEnvironment)
```
agramfort: @papadop  this happens otherwise:

https://travis-ci.org/openmeeg/openmeeg/jobs/130379872

agramfort:
friendly reminder @papadop :)

OpenMEEG/CMakeLists.txt
papadop: Obviously not the proper fix.
I'm committing the following instead:

     #   Hdf5 dir
 
     set(TestConfigFile "${CMAKE_BINARY_DIR}/TestConfig.cmake")
-    get_property(HDF5_LIB TARGET hdf5 PROPERTY LOCATION)
-    get_filename_component(HDF5_LIB ${HDF5_LIB} DIRECTORY)
-    get_filename_component(HDF5_ROOT_DIR ${HDF5_LIB} DIRECTORY)
+
+    set(HDF5_ROOT_DIR ${HDF5_LIBRARY_DIRS})
+    if (TARGET hdf5)
+        get_property(HDF5_LIB TARGET hdf5 PROPERTY LOCATION)
+        get_filename_component(HDF5_LIB ${HDF5_LIB} DIRECTORY)
+        get_filename_component(HDF5_ROOT_DIR ${HDF5_LIB} DIRECTORY)
+    endif()

Can you merge (from master) and check this works ??
Out of curiosity, did the tests work. This is made for tests to find the proper libraries. But I think that for windows and mac I'm installing everything in the same install tree, so all of this might be totally obsolete.... For the time being I'm  being conservative....

agramfort:
let's see if travis is happy

cc @papadop

agramfort:
@papadop I get IO errors when building on mac without the super project. Eg. https://travis-ci.org/openmeeg/openmeeg/jobs/132068911 Any idea?

papadop:
This is probably related to my patch for converting matio errors to exceptions.... But I cannot see why like this.... Which version of matio is installed on the system ? It might be that it is too old and that something is incompatible (even though that seems strange as I think the error facility in matio is rather old).

build_tools/travis/install.sh
```diff
@@ -25,39 +28,88 @@ else
...
 cd build
 
 # XXX : BUILD_SHARED should be used to set global defaults
 
 if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
-  cmake \
-      -DATLAS_INCLUDE_PATH:PATH=/usr/include/atlas \
+  if [[ "$USE_PROJECT" == "0" ]]; then
+      # Install MATIO
+      wget https://github.com/openmeeg/matio-openmeeg/archive/master.zip
```
agramfort: @papadop look here. I take current master

agramfort:
See diff and my inline comment

agramfort:
@papadop I can replicate the travis error on my debian system. Config with:

```
cmake -DBUILD_SHARED:BOOL=ON -DBUILD_DOCUMENTATION:BOOL=OFF -DBUILD_TESTING:BOOL=ON -DENABLE_PYTHON:BOOL=OFF -DENABLE_PACKAGING:BOOL=ON -DUSE_ATLAS:BOOL=ON -DUSE_SYSTEM_matio:BOOL=ON -DUSE_SYSTEM_hdf5:BOOL=ON -DUSE_SYSTEM_zlib:BOOL=ON -DCMAKE_SKIP_RPATH:BOOL=OFF -DBUILD_SHARED_LIBS_OpenMEEG:BOOL=ON -DBUILD_SHARED_LIBS_hdf5:BOOL=ON -DBUILD_SHARED_LIBS_matio:BOOL=ON -DBUILD_SHARED_LIBS_zlib:BOOL=ON ../../src/openmeeg
```

I get:

```
[ 16%] Linking CXX executable OpenMEEGMathsTest-vector
cd /cal/homes/gramfort/work/bin/openmeeg/OpenMEEG/build/libs/OpenMEEGMaths/tests && /cal/homes/gramfort/local/bin/cmake -E cmake_link_script CMakeFiles/OpenMEEGMathsTest-vector.dir/link.txt --verbose=1
/usr/bin/c++    -std=c++11 -O3 -DNDEBUG   CMakeFiles/OpenMEEGMathsTest-vector.dir/vector.cpp.o  -o OpenMEEGMathsTest-vector -rdynamic ../src/libOpenMEEGMaths.so.1.1.0 /usr/lib/atlas-base/libatlas.so /usr/lib/atlas-base/liblapack_atlas.so /usr/lib/atlas-base/libcblas.so -Wl,-rpath,/cal/homes/gramfort/work/bin/openmeeg/OpenMEEG/build/libs/OpenMEEGMaths/src:/usr/lib/atlas-base
CMakeFiles/OpenMEEGMathsTest-vector.dir/vector.cpp.o: In function `main':
vector.cpp:(.text.startup+0x9d3): undefined reference to `dlange_'
../src/libOpenMEEGMaths.so.1.1.0: undefined reference to `dsptrf_'
../src/libOpenMEEGMaths.so.1.1.0: undefined reference to `dsptrs_'
../src/libOpenMEEGMaths.so.1.1.0: undefined reference to `dgesdd_'
````

@eolivi any idea? thanks for any help.

eolivi:
I have the same problem too on my debian. 
I've tried all the atlas packet configurations... but still.
(for now I can only compile without ATLAS).
I'll try to re-investigate this, but I'll prefer spending time on switching to Openblas

agramfort:
I feel less alone :)

at least it works on @papdop 's machine :) :)

papadop:
Well. ATLAS has probably somewhat rotted. I'll look at it next week (currrently abroad). Nevertheless ATLAS was broken recently (a bug in the lib). OpenBlas (or another name which is the newest) is supported as far as I remember (I did that as a countermeasure for ATLAS). I'll verify all of this next week.

agramfort:
openblas is not in master. What was commited was not completed.

agramfort:
ok travis is happy.

to sum up. Currently what does not work is:

- use system MATIO on linux
- travis builds without the super project (tests don't pass)

I'll do this in a next PR.

PR #113, opened 2016-05-15
add appveyor.yml

agramfort:


agramfort:
@papadop so far no success with appveyor. See:

https://ci.appveyor.com/project/papadop/openmeeg/build/1.0.61#L153

any idea what going on?

maybe @dang-kai you have an idea?

tguillemot:
Ok there are several problems with CMake and some external libraries to solve before adding appveyor.

For the external libraries, I've propose those two PR :
- for `matio` https://github.com/openmeeg/matio-openmeeg/pull/5
- for `hdf5` https://github.com/openmeeg/hdf5-matio/pull/1
I've put the websites which explained the solutions on the respective PR.

For the CMake file, the main problem comes from the attribute `SOURCE_DIR` of `ExternalProject_Add`. In fact as the `SOURCE_DIR` is also the same as the one use to put the install file, ... everything is removed when they decompress which is downloaded creating the issue.
To solve it the best way is to modify the `Ep_Initialisation.cmake` file and changing the SOURCE_DIR to `src`. Once this is solved, there are other problems. 

I want to let you my progress but I will continue to investigate and I will propose a solution when it's done.

agramfort:
sounds like you have a plan. I'll look at the diff. thx !

PR #116, opened 2016-06-02
correct email adresses + for a VTK tool: solve seg_fault problem, and improve tool.

eolivi:


OpenMEEG/tools/vtp_to_meshes.cpp
```diff
@@ -49,17 +49,17 @@ int main( int argc, char **argv) {
 
     print_version(argv[0]);
 
-    command_usage("Convert meshes into a single VTP file.");
-    std::string input;
-    std::string output;
+    command_usage("Convert a single VTK/VTP into meshes.");
+    const char * input;
+    const char * output;
```
agramfort: why this change?

OpenMEEG/tools/meshes_to_vtp.cpp
```diff
@@ -49,50 +49,59 @@ int main( int argc, char **argv) {
...
-    input[5] = command_option("-i6", (const char *) NULL, "Input Mesh");
-    input[6] = command_option("-i7", (const char *) NULL, "Input Mesh");
-    name[0]  = command_option("-n1", (const char *) "1", "Mesh Name");
-    name[1]  = command_option("-n2", (const char *) "2", "Mesh Name");
-    name[2]  = command_option("-n3", (const char *) "3", "Mesh Name");
-    name[3]  = command_option("-n4", (const char *) "4", "Mesh Name");
-    name[4]  = command_option("-n5", (const char *) "5", "Mesh Name");
-    name[5]  = command_option("-n6", (const char *) "6", "Mesh Name");
-    name[6]  = command_option("-n7", (const char *) "7", "Mesh Name");
+    geom_filename = command_option("-g", (const char *) NULL, "Input geometry");
```
agramfort: can you explain to me the logic for the change? just to follow

OpenMEEG/tools/vtp_to_meshes.cpp
eolivi: -This "Convert meshes into a single VTP file"  was not the correct message.
- string Vs char * : command_option function requires "char *" and not "string". This worked somehow, but cause a seg_fault in certain cases (try "om_vtp_to_meshes -h" for ex.)

OpenMEEG/tools/meshes_to_vtp.cpp
eolivi: nothing changed for meshes.
But I added a way so that the user can (instead of providing each mesh separately), specify the geometry file (where meshes are already there and named) to convert it into a single Vtk/vtp file

agramfort:
great thanks @eolivi

PR #119, opened 2016-07-01
Openblas

eolivi:
provide support for OpenBlas on linux. (option USE_OPENBLAS).


eolivi:
Hello !
Here is my PR for Openblas support.
- please try this at home.
- the same has to be done for MAC and windows... if needed
- if it is ok for you: then we need to tell the user to use openblas instead of atlas (README.md), and also tell travis to use it.

agramfort:
cool !

@eolivi can I ask you to see if you can add this build option to travis ? it would be great to have continuous integration for this.

agramfort:
@eolivi you need to uncomment:

-    # - os: linux
-    #   compiler: gcc
-    #   env:
-    #     - USE_SYSTEM=1
-    #     - USE_PROJECT=0

otherwise it is not tested. It's currently uncommented as it's broken with ATLAS on Linux :(

cc @mclerc @papadop

eolivi:
the error is not due to openblas (even if it complains at first because it could not locate the OpenBLASConfig.cmake, after it finds the lib "Openblas found").
The pb is regarding files writing...
1: terminate called after throwing an instance of 'OpenMEEG::maths::BadFileOpening'
1: what(): Exception: Unable to open the file tmp_matrix.mat for writing

agramfort:
@eolivi your git history is messy. I can live with that but in the future never do:

git merge master

in your PR branch. If you need to update a branch use git rebase.

that being said does it seem to work with openblas on travis?

eolivi:
ok (thanks for advices @agramfort ), please do not merge this PR yet, I am trying to see if I can understand and fix travis failures.
I'll present a cleaner PR after.

agramfort:
why did you close?

PR #120, opened 2016-07-02
STY: fix clang warnings

agramfort:


agramfort:
thx for the fast reaction !

PR #121, opened 2016-07-11
support for openblas on linux.

eolivi:
it works on linux (correctly linked to libopenblas.so and faster than standard lapack)
But I wish @papadop could have a look to OpenMEEGMathsConfig.h to tell whether the support is done optimally.
Because it seems I'm only linking with libopenblas.so and using blas and lapack signatures, and maybe there is a better way to do it.


agramfort:
closing in favor of #128

PR #123, opened 2016-07-12
[MRG] Add Appveyor

tguillemot:
I cherry pick the work of @agramfort #113 and correct it to make appveyor available.


appveyor.yml
```diff
@@ -0,0 +1,43 @@
...
+# scripts to run after build
+after_build:
+  - cd %P%
+  - 7z a c:/projects/openmeeg/openmeeg.zip *.exe -tzip
+  - cd c:/projects/openmeeg/OpenMEEG/build
+  - ctest -V
+
+artifacts:
+  - path: openmeeg.zip
+    name: openmeeg.zip
```
agramfort: does it properly generate the artifacts?

I suspect you need to call cpack or make package to generate the windows installers.

what would be awesome would be to update the builts from master branch to a download website. Don't know how to do this...

we would probably need an ssh key to upload somewhere...

agramfort:
@papadop any objection?

@tguillemot apparently artifacts are not there:

https://ci.appveyor.com/project/papadop/openmeeg/build/1.0.110/artifacts

tguillemot:
@agramfort As I said to you, I'm clearly not an expert of cmake and appveyor (as it's more or less the first time I do it), so it's totally possible that I've missed something. 

Concerning the artifacts, I totally missed that indeed. I've try to add it correctly with the new push.

Concerning deployment, there is an option with appveyor for that they propose : personnal web site, azure cloud service, github (for mor info [look at the doc](https://www.appveyor.com/docs/appveyor-yml) ). Just tell me where you want to push the artifacts and I will try to solve that.

tguillemot:
Ok there is an issue with cpack on appveyor :

tguillemot:
@agramfort Artifacts are here now and working on my computer.
There is still a warning during the installation concerning the PATH size. 
I'll see if I can solve it.

agramfort:
cool

can you also generate package as TGZ

cpack -G TGZ

should do it.

that's what is used for deployment by the matlab toolboxes that use this code.

agramfort:
thanks a ton @tguillemot !!!

🍻  🍻 🍻 🍻 🍻 🍻  !!!

tguillemot:
Youpi !!!

papadop:
Thank you for this work...

Unfortunately, the merge of PR #123 breaks the build in empty directory (but a source tree which already has the external libraries)... Seems to be related to the ${ep}_SOURCE_DIR change because URLs are not defined. Going back to 16b03f8dd3b93a026e546f31251d4e4f5017515a solves it. Will investigate but if someone beats me... 

Having external sources (and not the build directory) in the source directory is a choice. Often, while developing, you make some changes to sources without looking too much where they are, and often also you do a rm -rf of the build dir. I'd like to keep sources in the source directory.... Is there a specific reason why you changed this ?

Also I seem to see a lot a changes which seem to be related to line endings (are those trailing spaces or windows line endings).  If the later, I'd like to avoid those. It was maybe for that that there was some git stuff (I do not remember if we removed or not this git magic).

papadop:
Oups forget about the windows line endings. I saw it was trailing spaces.

PR #125, opened 2016-07-18
Corticalmat

eolivi:
solve a little bug introduced lately on corticalMat2, and add a proper test for it.


agramfort:
LGTM

merge if you're done

thx @eolivi !

PR #126, opened 2016-07-28
Sources in source dir

papadop:
Put back external sources in the source directory.

This is to avoid a complicated situation if someone wants to change an external source. If sources are in the build dir, then they are local to that build and propagation between multiple builds might be a nightmare. Also a simple rm -rf of the build dir (which is IMHO a reasonable operation) will make you loose all your changes. One possibility would be to make the build src dirs read only, but then changing something in those external sources would be quite tedious.


papadop:
Ah, I see this is interacting badly with AppVeyor. @tguillemot, can you elaborate on why ??

agramfort:
@papadop can you avoid pushing the dev branches to the main repo?

plus push to your github fork

agramfort:
@papadop appveyor is not happy :(

papadop:
What is the problem in pushing branches to the main repo ??? In this way, everyone can see them easily (without having to add yet another remote). Branches are exactly made for that, no ?

papadop:
And I do not see why this should break appveyor... 
That is why I asked Thierry with details.

papadop:
Seems (at least partly) related to this message:
Creating "x64\Release\zlib\zlib.tlog\unsuccessfulbuild" because "AlwaysCreate" was specified
Should we remove these source directories before every build.
Not a perfect solution, but an easy one ?

papadop:
Saw that the build was in source directory. Modified that to see whether it improves things.

agramfort:
> What is the problem in pushing branches to the main repo ???

when someone does a git clone of upstream repo he will get all your dev branches rather than just the clean master and the release branches

> In this way,
> everyone can see them easily (without having to add yet another remote).

that's what pull requests are for. You see a clean diff even if PR comes from
another remote.

> Branches are exactly made for that, no ?

I would say no. It's branches + remotes + github PR than make it work not branches alone :)

papadop:
But PR just allow you to look at changes not to test/debug changes... While I can understand that for long standing branches, I may agree, for short lived ones (as I expect this one to be), it seems much more easy to just use branches. Having multiple remotes and branches is a path to troubles (at least for me, with my limited fu on git).

papadop:
Anyhow, I hope that my last patch will fix the appveyor build. I suppose the travis problem is a transient one, as I made strictly no changes outside of appveyor.xml in the last patches.

agramfort:
@papadop you can add to your openmeeg/.git/config file

```
[remote "origin"]
	url = git@github.com:openmeeg/openmeeg.git
	fetch = +refs/heads/*:refs/remotes/origin/*
	fetch = +refs/pull/*/head:refs/remotes/origin/pr/*
```

and

git fetch origin

will fetch all the PRs for you. Then you can do:

git checkout pr/126

and you're done. No need to add a remote :)

agramfort:
appveyor is not happy yeta

PR #132, opened 2016-07-29
attempt to fix appveyor artifacts

agramfort:
@papadop you broken the artifacts creation on appveyor

trying to fix it here.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,7 +32,7 @@ before_test:
   - ctest -V
 
 artifacts:
-  - path: OpenMEEG-*-x64--static.exe
+  - path: c:\projects\openmeeg\build\OpenMEEG-*-x64--static.exe
     name: OpenMEEG-x64.exe
-  - path: OpenMEEG-*-x64--static.tar.gz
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -32,7 +32,7 @@ before_test:
   - ctest -V
 
 artifacts:
-  - path: OpenMEEG-*-x64--static.exe
+  - path: c:\projects\openmeeg\build\OpenMEEG-*-x64--static.exe
```
tguillemot: First problem : appveyor artifacts doesn't support the complete path. So you have to change it as it was before

tguillemot: l24-27 seems strange to me. 
Normally, cpack has to be called in the directory `c:\projects\openmeeg\build\`, it seems the case but `move` move the files in the same place. If you change the . by c:/ I think it will work.

tguillemot:
@agramfort I think I've found the problem I will try on my repo to be sure.

appveyor.yml
agramfort: please open a new PR. I'll close mine

tguillemot:
@agramfort You can close this PR and see #135.

PR #134, opened 2016-07-31
attempt to setup automatic deployment to gforge of packages

agramfort:


agramfort:
won't work with PR from remote fork as env vars are not exposed for security reasons

PR #136, opened 2016-08-01
[MRG] Unilapack

eolivi:
complete support of openblas
this is a rebase from https://github.com/papadop/openmeeg/tree/UniLapack
+fixing conflicts


eolivi:
after the "make it works", this is the "make it nice" and "make it fast" , tell me if it works for MAC @agramfort  please.
@papadop

agramfort:
travis is not happy :(

see eg : https://travis-ci.org/openmeeg/openmeeg/jobs/148830968

see if it's not a cmake config option you're mission in travis.yml file if it works on your machine.

eolivi:
What do we set as default Lapack implementation on Linux ? 
- ATLAS (but broken)
- OpenBlas ?
- standard Lapack ? (I've set this one)
@agramfort @papadop

agramfort:
Current default is to use the shipped lapack but it should work with system atlas too

eolivi:
ok, but system atlas is broken still...

agramfort:
I know... don't know why :(

@papadop works your machine?

if so when is your birthday
http://www.redbubble.com/fr/people/ramiro/works/17472964-shrug-it-works-on-my-machine-programmer-excuse-design-white-on-black
? :-)

OpenMEEG/cmake/UseACML.cmake
```diff
@@ -0,0 +1,5 @@
+if (USE_ACML)
+    set(ACML_ROOT "/home/papadop/Downloads/acml/gfortran64")
```
eolivi: ooops TODO .. @papadop

papadop: ACML support has been long ago suspended. I probably kept the code in case we revive this port one day. For now, I do not think it can be activated. But if you can reactivated the support. I'd be happy. That being said, this is also why I support using flens because I think they support ACML, MKL, Atlas, OpenBLAS and maybe more...

agramfort: I vote for getting rid of any ACML reference in our code as nobody ever used it in 10 years. If we go with flens we'll have it for free. Until then let's avoid maintaining lines of code for no benefit.

agramfort:
@eolivi let me know if you need another pair of eyes to review and if not when travis is happy.

eolivi:
@agramfort and @papadop now you can start reviewing as it is (finally) green on travis and windows :-D
Do you think we keep the option BLASLAPACKIMPLEMENTATION , I think it is too long, but maybe not very important since we have the "Auto" possibility by default now. or we could rename it into "LAPACK_IMPL" ?

OpenMEEG/libs/OpenMEEG/src/assembleSensors.cpp
```diff
@@ -117,6 +117,7 @@ namespace OpenMEEG {
         unsigned p0_p1_size = (geo.size() - geo.nb_current_barrier_triangles());
 
         Matrix FergusonMat(3*nbIntegrationPoints, geo.nb_vertices());
+        FergusonMat.set(0.0);
```
agramfort: is that a bug fix?

build_tools/travis/before_install.sh
```diff
@@ -35,43 +36,41 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
 else
     # Install some custom requirements on Linux
     # g++4.8.1
-    if [ "$CXX" == "g++" ]; then sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test; fi
+    sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test; 
+    sudo apt-get install -qq g++-4.8;
+    export CXX="g++-4.8";
```
agramfort: why not protecting these lines in the else of the "if [ "$CXX" == "clang++" ]; then"

agramfort:
2 small comments:
- I see a few debug comments line that should be removed once we're happy here (meaning CIs are happy)
- I don't like much the mix of snake and camel case in new import names + the redundant presence of OM in the name. For example OM_Atlas_Config.h for me should be AtlasConfig.h or maybe MatLibAtlasConfig.h

papadop:
For the last comment, I beg to somewhat differ. Certainly not AtlasConfig.h that is too generic and can be used by someone else. MatLibAtlasConfig.h is slightly better even though it is still somewhat generic. User are not exposed to these files (i.e. they are included for them), we have all the latitude to obfuscate their names to avoid a potential conflict with another package (which would be quite difficult to diagnose). In this sense, the OM prefix is not that bad. This kind of prefix is often used for typical files like config.h... Now if you prefer OpenMEEGMatLibAtlasConfig.h why not...

agramfort:
ok then +1 for OpenMEEGMatLibAtlasConfig.h

OpenMEEG/libs/OpenMEEG/src/assembleSensors.cpp
eolivi: yes it is ! at least on win10... but I don't know why it was never spotted before.... 
nevertheless, usually, at other places when we declare a matrix, we set it to zero.

agramfort:
ok... do we have any idea of how critical this bug fix is? if it generated
really wrong results in the past for our users?

build_tools/travis/before_install.sh
eolivi: **anyway I'll reformat that file soon,** but as on travis the ubuntu version is old (trusty is 14.04), the libopenblas is bugged (not all the files as openblas_config.h), so I need the 
sudo sed -i -e 's/trusty/vivid/g' /etc/apt/sources.list
and when it was built on travis, I had a clang build failed whereas the g++ was ok, and this because of libopenblas .... there should not be any reason for that, but adding the g++ ppa, solved this...

eolivi:
so I really go for this ?
OM_Atlas_Config.h -> OpenMEEGMathsAtlasConfig.h
OM_BlasLapack_Config.h -> OpenMEEGMathsBlasLapackConfig.h
OM_C_BlasLapack.h -> OpenMEEGMathsCBlasLapack.h
OM_F_BlasLapack1.h -> OpenMEEGMathsFBlasLapack1.h
OM_F_BlasLapack.h -> OpenMEEGMathsFBlasLapack.h
OM_MKL_Config.h -> OpenMEEGMathsMKLConfig.h
OM_OpenBLAS_Config.h -> OpenMEEGMathsOpenBLASConfig.h
OM_vecLib_Config.h -> OpenMEEGMathsvecLibConfig.h

by the way if I was mixing the cases, it was because the different projects (MKL, Atlas, ...) were called that way with that case. I thought that I'll have troubles with find_package of cmake, but actually, it should be case-insensitive... I can uniformize

eolivi:
regarding the bug, it was really wrong (matrix filled with ??E+66), this happen only since this PR... but I did not investigate deeply on why this never happened before.. setting the iniatial fergusson matrix to 0. seems legit enough for me.

papadop:
OK for the renaming. But indeed cmake is not case insensitive verytime and that is specifically true in the case of find_package. So be careful...

eolivi:
I can check again, but that was probably the reason I decided to go for following the original package name case

build_tools/travis/before_install.sh
agramfort: ok thx. Maybe add a comment to explain why it's needed?

agramfort:
ok for renaming as suggested

OpenMEEG/cmake/BlasLapackOption.cmake
```diff
@@ -0,0 +1,45 @@
...
+set(MKL      USE_MKL)
+set(LAPACK   USE_LAPACK)
+set(vecLib   USE_VECLIB)
+set(Auto     USE_AUTO)
+
+# the default case
+if (BLASLAPACK_IMPLEMENTATION)
+    set(BLASLAPACK_IMPLEMENTATION_DEFAULT ${BLASLAPACK_IMPLEMENTATION})
+else()
+    if (WIN32)
```
papadop: The two cases of the else are equal. Simplify.

OpenMEEG/cmake/UseAtlas.cmake
```diff
@@ -1,148 +1,55 @@
...
-    option(USE_ATLAS "Build the project using ATLAS" OFF)
-    option(USE_MKL "Build the project with MKL" OFF)
-    if (NOT BUILD_SHARED_LIBS)
-        set(CMAKE_FIND_LIBRARY_SUFFIXES ".a;.so")
-        if (APPLE)
-            set(CMAKE_FIND_LIBRARY_SUFFIXES "${CMAKE_FIND_LIBRARY_SUFFIXES};.dylib")
-        endif()
-    endif()
-endif()
+#   Create a FindAtlas.cmake ?
```
papadop: +1

OpenMEEG/cmake/UseAuto.cmake
```diff
@@ -0,0 +1,27 @@
+
+if (USE_AUTO)
+    # this macro tries to use library in the defined priority
+    macro(priorities)
+        foreach (lib ${ARGN})
+            string(TOUPPER ${lib} LIB)
+            set(USE_${LIB} true)
+            include(Use${lib})
```
papadop: I'm slightly worried by the fact that this macro may import some variables from the USEXXX cmake files in the global scope which may create problems difficult to spot. Is there a way to transform this macro into a function and export in the global scope only the variables that we know to need.

OpenMEEG/libs/OpenMEEG/include/integrator.h
```diff
@@ -177,7 +177,7 @@ namespace OpenMEEG {
 
     static const unsigned nbPts[4] = {3, 6, 7, 16};
 
-    template <class T, class I>
```
papadop: I do not understand why the renaming is useful. Not a reason to block the merge though. Note also that "class" should be replace with typename one day or the other.

OpenMEEG/libs/OpenMEEG/src/assembleSensors.cpp
papadop: Strange but the correction seems OK. Probably before, it was using a portion of memory that was never used by the system (and thus set to zero). Somehow, now this reuses some memory already used hence the poor initialisation.

papadop:
OK for me. You can merge when you have made the changes alluded in the comments.

OpenMEEG/cmake/UseAuto.cmake
eolivi: done

OpenMEEG/libs/OpenMEEG/include/integrator.h
eolivi: actually it is because when using openblas, complex.h is included, so the compiler complains (because I is not a complex imaginary...) hence the fix.
I've updated with typename this time

eolivi: actually I will remove that commit as there is a much better fix

agramfort:
ok

please set title of the PR to "[MRG] Unilapack" when it's ready for merge.

eolivi:
@agramfort ok for you ?

OpenMEEG/cmake/UseMKL.cmake
```diff
@@ -0,0 +1,13 @@
+
+if (USE_MKL)
+    find_package(MKL ${REQUIRED}) # e.g REQUIRED can be REQUIRED or QUIET
+    if (MKL_FOUND)
+        include_directories(${MKL_INCLUDE_DIR})
+        set(LAPACK_LIBRARIES ${MKL_LIBRARIES})
+        get_filename_component(MKL_DLL_DIR "${MKL_LIBRARIES}" DIRECTORY) 
+        #message(${LAPACK_LIBRARIES}) # for debug
```
agramfort: to remove?

OpenMEEG/cmake/UseOpenBLAS.cmake
```diff
@@ -0,0 +1,9 @@
+
+
```
agramfort: remove these empty lines

OpenMEEG/cmake/UsevecLib.cmake
```diff
@@ -0,0 +1,12 @@
+
```
agramfort: same here

OpenMEEG/libs/OpenMEEGMaths/include/BlasLapackImplementations/OpenMEEGMathsCBlasLapack.h
```diff
@@ -0,0 +1,28 @@
...
+        #define DGETRF(X1,X2,X3,X4,X5,X6) LAPACK(dgetrf,DGETRF)(&X1,&X2,X3,&X4,X5,&X6)
+        #define DGETRI(X1,X2,X3,X4,X5,X6,X7) LAPACK(dgetri,DGETRI)(&X1,X2,&X3,X4,X5,&X6,&X7)
+    #else
+        #define DGETRF(X1,X2,X3,X4,X5) LAPACK(dgetrf,DGETRF)(CblasColMajor,X1,X2,X3,X4,X5)
+        #define DGETRI(X1,X2,X3,X4)    LAPACK(dgetri,DGETRI)(CblasColMajor,X1,X2,X3,X4)
+    #endif
+#else
+    #define DGETRF LAPACK(dgetrf,DGETRF)
+    #define DGETRI LAPACK(dgetri,DGETRI)
+#endif
```
agramfort: to remove or uncomment as if 0?

build_tools/travis/before_install.sh
agramfort: can you add the comment?

build_tools/travis/before_install.sh
```diff
@@ -35,43 +36,41 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
...
+    # # Handle MATIO
+    # sudo apt-get update -qq
+    # # to prevent IPv6 being used for APT
+    # sudo bash -c "echo 'Acquire::ForceIPv4 \"true\";' > /etc/apt/apt.conf.d/99force-ipv4"
+    # # The ultimate one-liner setup for NeuroDebian repository
+    # bash <(wget -q -O- http://neuro.debian.net/_files/neurodebian-travis.sh)
+    # # But we actually want -devel repository just to get matio backport
+    # sudo sed -ie 's,neuro.debian.net/debian ,neuro.debian.net/debian-devel ,g' /etc/apt/sources.list.d/neurodebian.sources.list
+    # # Just to get information about available versions
+    # apt-cache policy libmatio-dev
```
agramfort: no need for neurodebian anymore?

eolivi:
@agramfort ok I've done the changes you wanted (almost all). I did a commit --amend, so it is hard for you to see only these few changes, I did that to group all commits related to this PR on unilapack. would you have prefered small commits each time ?

agramfort:
ok for me.

yes I prefer small incremental commits so I just need to read tiny updates

so no more use of neurodebian?

agramfort:
thx @eolivi

Issue #139, opened 2016-10-15
Build fails on OSX 10.10


dpo:
On OSX, the Accelerate framework provides the necessary BLAS and LAPACK implementation, including `cblas.h` and `clapack.h`. It seems `openmeeg` only looks for ATLAS or MKL. I have to set `-DUSE_ATLAS=OFF` and `-DCMAKE_C_FLAGS=-framework Accelerate` to get CMake to find `cblas.h` and `clapack.h`. However, I then hit the build error:

```
/tmp/openmeeg-20161015-82762-q9osu7/openmeeg-release-2.1/libs/MatLib/MatLib/MatLibConfig.h:54:18: fatal error: 'FC.h' file not found
        #include <FC.h>
                 ^
```

agramfort:
code should work on mac. Travis is normally testing it. See:

https://github.com/openmeeg/openmeeg/blob/master/.travis.yml
https://github.com/openmeeg/openmeeg/blob/master/build_tools/travis/before_install.sh
https://github.com/openmeeg/openmeeg/blob/master/build_tools/travis/install.sh

does it help? 

basically we install stuff with homebrew
then running a long cmake command with necessary options.

does it help?

ilovezfs:
@agramfort when building for macOS, your Travis configuration appears to use OpenBLAS. @dpo above is talking about using the Accelerate framework, which ships natively with macOS.

agramfort:
@dpo can you paste the full line you use with cmake

cmake -DCMAKE_C_FLAGS=-framework Accelerate ...

?

I can then try on my mac.

ilovezfs:
@agramfort basically anything depending on `veclibfort` should be a useful example for you by the way
https://gist.github.com/ilovezfs/3977538f7c847b6c8f99d8686262c07b

dpo:
@agramfort Sorry for the delay. Here's my complete CMake command:
```
[...]

PR #140, opened 2016-10-29
add auto detection of veclib + update travis

agramfort:
follow up on #139


agramfort:
CIs are happy merging

PR #141, opened 2016-11-05
use of libmatio with hdf5 support from brew on mac os x

agramfort:
use libmatio with hdf5 support from brew on mac os x on travis


agramfort:
travis is happy

merging

PR #150, opened 2017-02-17
Cleancode

eolivi:
and avoid dangling else problem.
cmake desambiguation on CMP0054


papadop:
It is almost impossible to see where is the important change...

agramfort:
@eolivi avoid cosmetic changes in the middle of real changes as indeed the diffs are not readable easily.

agramfort:
@eolivi are these only cosmetic changes?

if so +1 for MRG

eolivi:
now yes, only cosmetic changes following the established coding_guidelines.txt ;-)

agramfort:
thx @eolivi

PR #153, opened 2017-02-20
[MRG] cosmit: replace class into typename

eolivi:
as @papadop advised


agramfort:
+1 for merge when CIs are happy

PR #155, opened 2017-02-20
[MRG] fix support for vecLib on Mac + restore this case on travis.

eolivi:
choosing BLASLAPACK_IMPLEMENTATION=Auto 
uses in priority OpenBLAS (on MAC as in linux).
so the vecLib, was not tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
       env:
         - USE_SYSTEM=1
         - USE_PROJECT=1
-        - BLASLAPACK_IMPLEMENTATION=OpenBLAS
+        - BLASLAPACK_IMPLEMENTATION=vecLib
     - os: osx
       compiler: clang
       env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -14,7 +14,7 @@ matrix:
       env:
         - USE_SYSTEM=1
         - USE_PROJECT=1
-        - BLASLAPACK_IMPLEMENTATION=OpenBLAS
+        - BLASLAPACK_IMPLEMENTATION=vecLib
```
agramfort: are you sure? what is the default implem for mac? the test config about uses default on mac. Should I understand that OpenBLAS is the default on mac now?

eolivi: yes openblas is default on Mac and unix

eolivi: one can change that very easily in the file OpenMEEG/cmake/UseAuto.cmake

agramfort:
ok thx

PR #156, opened 2017-02-21
[MRG] fixing the pseudo inverse test.

eolivi:
- correct failing test due to non-initialized memory
- use a sparseMatrix to store the diagonal S in the svd.
- return the transposed V in the svd

### DIFF
--- a/OpenMEEG/libs/OpenMEEGMaths/tests/full.cpp
+++ b/OpenMEEG/libs/OpenMEEGMaths/tests/full.cpp
@@ -42,6 +42,7 @@ knowledge of the CeCILL-B license and that you accept its terms.
 
 #include <OpenMEEGMathsConfig.h>
 #include <matrix.h>
+#include <sparse_matrix.h>
 #include <generic_test.hpp>
 
 int main () {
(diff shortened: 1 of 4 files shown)


agramfort:
can you put back the test so we can see if it fixes the pb?

OpenMEEG/libs/OpenMEEGMaths/tests/full.cpp
```diff
@@ -108,33 +109,38 @@ int main () {
...
+        std::cout << "U :" << std::endl;
+        U.info();
+        std::cout << "S :" << std::endl;
+        S.info();
+        std::cout << "W :" << std::endl;
+        W.info();
         std::cerr << "Error: SVD is WRONG-1" << std::endl;
         exit(1);
     }
-    /* temporarly removing the pseudo inverse test, as casuing troubles on travis
```
eolivi: I remove the "/*" and "*/"so the test is back on

OpenMEEG/libs/OpenMEEGMaths/tests/full.cpp
```diff
@@ -108,33 +109,38 @@ int main () {
...
+    zero = M1*M1pinv*M1-M1;
     if (zero.frobenius_norm()>eps) {
+        std::cout << "pinverse = M^{+} = " << std::endl;
+        M1pinv.info();
+        std::cout << "M*M^{+}*M-M = zero matrix = " << std::endl;
         zero.info();
         std::cerr << "Error: PseudoInverse is WRONG-2" << std::endl;
         exit(1);
     }
-    */
```
eolivi: end here

agramfort:
awesome ! thx @eolivi

PR #157, opened 2017-02-22
[MRG] restore the BLASLAPACK_IMPLEMENTATION option string correctly

eolivi:


agramfort:
@eolivi merge if you're done. thx

PR #159, opened 2017-02-23
MKL fix lapack implementation

eolivi:


OpenMEEG/libs/OpenMEEGMaths/src/matrix.cpp
```diff
@@ -142,14 +142,14 @@ namespace OpenMEEG {
...
-    #else
-        std::cerr<<"svd not implemented without blas/lapack"<<std::endl;
-    #endif
         if ( Info < 0) {
             std::cout << "in svd: the "<< -Info << "-th argument had an illegal value." << std::endl;
         } else if ( Info > 0) {
             std::cout << "in svd: DBDSDC did not converge, updating process failed." << std::endl;
         }
+    #else
+        std::cerr<<"svd not implemented without blas/lapack"<<std::endl;
```
agramfort: this is no SVD in MKL? I would be surprised.

eolivi: no look again, that was just to fix a warning, regarding Info variable being checked without being initialized. the #else was not properly located. (nothing to do with MKL)

OpenMEEG/libs/OpenMEEGMaths/include/matrix.h
```diff
@@ -187,7 +187,7 @@ namespace OpenMEEG {
     }
     
     inline double Matrix::frobenius_norm() const {
-    #ifdef HAVE_LAPACK
+    #if defined(HAVE_LAPACK) && !defined(NO_DLANGE)
```
eolivi: I just had difficulties finding the dlange in the mkl, but we only use it to get the frobenius norm in very few cases (mostly tests). so I disabled the use of dlange and instead use a for loop.

agramfort: why not use dnrm2 then?

http://www.netlib.org/lapack/explore-3.1.1-html/dnrm2.f.html

eolivi: lapacke_dlange is supported on new versions of MKL, but I cannot test it, because the machines on ci are only with 2013 version of MKL, so that was a fix to be able to use MKL not depending on the version.
And as I told you it is rarely used the frobenius norm (only in tests). I'll try to get a newer version of MKL to remove this workaround.

agramfort:
ok I get this but I am saying let's stick to the old version of MKL so we
don't depend on a bleeding edge version but let's use dnrm2 in our code so
DLANGE is not used anywhere and we are not using a slower code when
switching to MKL. Makes sense?

OpenMEEG/libs/OpenMEEGMaths/include/BlasLapackImplementations/OpenMEEGMathsMKLConfig.h
```diff
@@ -10,7 +10,12 @@
 
 #define CLAPACK_INTERFACE
 
-#define DLANGE(X1,X2,X3,X4,X5,X6)       LAPACK(dlange,DLANGE)(LAPACK_COL_MAJOR,X1,X2,X3,X4,X5)
+
+extern "C" {
+    double dlange(const char&,const int&,const int&,double*,const int&,double*);
+}
+
+#define DLANGE dlange
```
eolivi: Actually dlange is in the MKL even in older versions (but not with the c++ header of lapacke), I had troubles to understand, but what we do for all other lapack functions is to call them directly through the c++ lapacke headers (mkl_lapacke.h), but there are no lapacke_dlange in MKL version 2013 (whereas there is such a signature in MKL 2017). 
so this fixes the problem, by calling the dlange provided (in all versions) in the mkl_lapack.h.
If you want, we can wait a look of @papadop for confirmation. but at least it now build on windows MKL 2013 AND uses the dlange lapack function.

agramfort:
great

thx a lot @eolivi 

merge if you're done here.

PR #167, opened 2017-03-17
minor, + fix warnings on Windows, update unused VTK module.

eolivi:
this is for solving many warning on Windows mainly.


cmake/externals/projects_modules/VTK.cmake
```diff
@@ -1,76 +0,0 @@
-################################################################################
```
eolivi: not a pb to remove that file ? was it ever used ?

agramfort: if tests pass no objection :)

agramfort:
no objection to merge if CIs are happy

cmake/externals/projects_modules/VTK.cmake
papadop: I has never been used (except with medInria). The idea was to allow people that cannot install things on their machine to still compile OM (in the same way we provide lapack, matio, hdf5, and zlib). I'm for keeping it and allow its use which should be extremely simple....

OpenMEEG/libs/OpenMEEG/src/mesh.cpp
```diff
@@ -942,7 +944,7 @@ namespace OpenMEEG {
         unsigned int mesh_step[1] = {0}; // mesh_step
         os.write((char*)mesh_step, sizeof(unsigned int));
 
-        unsigned int vertex_number[1] = {nb_vertices()}; // vertex number
+        unsigned int vertex_number[1] = {(unsigned)nb_vertices()}; // vertex number
```
papadop: Use static_cast instead of C casts....

OpenMEEG/libs/OpenMEEGMaths/include/matrix.h
```diff
@@ -222,7 +221,7 @@ namespace OpenMEEG {
         om_assert (istart+isize<=nlin() && jstart+jsize<=ncol());
 
         Matrix a(isize,jsize);
-        const int sz = static_cast<int>(isize);
+        const int sz = (int)isize;
```
papadop: NO !!

agramfort: :) can you educate us? :)

OpenMEEG/libs/OpenMEEGMaths/include/symmatrix.h
```diff
@@ -324,11 +324,11 @@ namespace OpenMEEG {
         // LU
         int *pivots = new int[nlin()];
         int Info = 0;
-        DSPTRF('U', nlin(), data(), pivots, Info);
+        DSPTRF('U', (int)nlin(), data(), pivots, Info);
         // Inverse
         int sz = (int) this->nlin() * 64;
         double *work = new double[sz];
-        DSPTRI('U', nlin(), data(), pivots, work, Info);
+        DSPTRI('U', (int)nlin(), data(), pivots, work, Info);
```
papadop: Not sure I like all those changes. First forget about C casting, it is not safe. Then OK, you silence the warning but silently introduce a bug. If you want to follow that path, you should test that the int(s) are not negative...

eolivi: max int in 32bits = 2'147'483'647
.... ok maybe we will have a matrix of 20'000 x 10'0000 one day...
I'll change that as we agreed into a function sizet_to_int

OpenMEEG/libs/OpenMEEGMaths/include/matrix.h
papadop: C casts are not type safe. C cast are mapped to 4 types in C++ static_cast, const_cast, dynamic_cast, and reinterpret_cast. dynamic_cast is of no interest here as we deal with POD. reinterpret_cast in C++ is the non type-safe cast (it is not relevant too here as it only deals with pointer and references). 

Quoting BS: C type cast should have been deprecated.... static_cast is the proper choice here. A bonus point if there is a check that the resulting number is not negative. To avoid very long lines, make a small inline function...

https://books.google.fr/books?id=EXfcAAAAQBAJ&pg=PA161&lpg=PA161&dq=C%2B%2B+C+type+casting+deprecated&source=bl&ots=2spW5eE_6W&sig=JCQDTqpEmAaEwkjP0yQvHISzPAM&hl=fr&sa=X&ved=0ahUKEwiegIXj9d3SAhWDHxoKHT6BB7gQ6AEITDAG#v=onepage&q=C%2B%2B%20C%20type%20casting%20deprecated&f=false

agramfort:
ok thanks.

OpenMEEG/libs/OpenMEEGMaths/include/linop.h
```diff
@@ -51,6 +52,14 @@ namespace OpenMEEG {
...
 
+    // to properly convert a size_t int to an int
+    inline OPENMEEGMATHS_EXPORT int sizet_to_int(const size_t& num)
+    {
+        int num_out = static_cast<int>(num);
+        om_assert(num_out >= 0);
+        return num_out;
+    }
+
     class OPENMEEGMATHS_EXPORT LinOpInfo {
```
eolivi: or maybe I do a template function to return an unsigned also ? (rarely needed... only in mesh.{h,cpp})

agramfort:
@eolivi let us know when CIs are happy

OpenMEEG/libs/OpenMEEG/src/mesh.cpp
```diff
@@ -975,7 +978,7 @@ namespace OpenMEEG {
         os.write((char*)normals_raw, sizeof(float)*nb_vertices()*3);       // normals
         unsigned char zero[1] = {0};
         os.write((char*)zero, sizeof(unsigned int));                       // arg size : 0
-        unsigned int faces_number[1] = {nb_triangles()};
+        unsigned int faces_number[1] = {(unsigned)nb_triangles()};
         os.write((char*)faces_number, sizeof(unsigned int));               // ntrgs
```
eolivi: I forgot that one. It will be in the next commit

OpenMEEG/libs/OpenMEEG/src/mesh.cpp
agramfort: Remains

OpenMEEG/libs/OpenMEEG/src/mesh.cpp
```diff
@@ -975,7 +977,7 @@ namespace OpenMEEG {
         os.write((char*)normals_raw, sizeof(float)*nb_vertices()*3);       // normals
         unsigned char zero[1] = {0};
         os.write((char*)zero, sizeof(unsigned int));                       // arg size : 0
-        unsigned int faces_number[1] = {nb_triangles()};
+        unsigned int faces_number[1] = {(unsigned)nb_triangles()};
```
agramfort: Here too

OpenMEEG/libs/OpenMEEGMaths/src/matrix.cpp
```diff
@@ -134,9 +134,9 @@ namespace OpenMEEG {
         int *iwork = new int[8*mini];
         int Info = 0;
         if ( complete ) { // complete SVD
-            DGESDD('A',nlin(),ncol(),cpy.data(),nlin(),s,U.data(),U.nlin(),V.data(),V.nlin(),work,lwork,iwork,Info);
+            DGESDD('A',(int)nlin(),(int)ncol(),cpy.data(),(int)nlin(),s,U.data(),(int)U.nlin(),V.data(),(int)V.nlin(),work,lwork,iwork,Info);
         } else { // only first min(m,n)
-            DGESDD('S',nlin(),ncol(),cpy.data(),nlin(),s,U.data(),U.nlin(),V.data(),V.nlin(),work,lwork,iwork,Info);
+            DGESDD('S',(int)nlin(),(int)ncol(),cpy.data(),(int)nlin(),s,U.data(),(int)U.nlin(),V.data(),(int)V.nlin(),work,lwork,iwork,Info);
```
agramfort: Static cast

agramfort:
That's it for me

OpenMEEG/libs/OpenMEEG/src/mesh.cpp
eolivi: yes ... first I used the same function sizet_to<unsigned> for that, but then on Windows, it created problems with __declspec(dllexport) ...  so I'll use only a static cast here (without a check). is it okay for you ?

OpenMEEG/libs/OpenMEEGMaths/src/matrix.cpp
eolivi: bravo !

agramfort:
thx @eolivi 

release now?

eolivi:
don't worry , Yes it will be very soon. I already could generate a package on windows with VTK + MKL (if you want I can put it online...).
I only have a few PR left to do before properly doing the release. (with the git tag, and maybe with python as well)

agramfort:
is the build of the packages automatic? eg using artifacts on appveyor? we
should aim for this so release is painless and release cicles can be short

eolivi:
you cannot build on appveyor with the MKL https://www.appveyor.com/docs/build-environment/#pre-installed-software 
or do you know a way ?
for automatic builds : ci.inria and jenkins ?

agramfort:
no idea :(

would openblas be an option on appveyor?

PR #168, opened 2017-03-17
Atlas fortran c

eolivi:
No need to check matio version anymore (since even the old trusty ubuntu have a correct version).
Finish with the FortranCinterface, that we keep for All blasLapack except OpenBLAS and MKL. (so FC.h create even on Mac)


agramfort:
no objection.

eolivi:
after @papadop ' comments.
please merge once tests finished

PR #169, opened 2017-03-22
Improving CodeCoverage adding a test on EIT. Removing a dupplicated gain.

eolivi:


agramfort:
@eolivi I see a segfault in EITSM-Head1 in appveyor

OpenMEEG/src/TestHead.txt
```diff
@@ -26,6 +26,7 @@ function(TESTHEAD HEADNUM)
     set(DIPPOS                 ${MODELBASE}.dip)
     set(DIPPOS-SKULLSCALP      ${MODELBASE}-skullscalp.dip)
     set(PATCHES                ${MODELBASE}.patches)
+    set(EITPATCHES             ${MODELBASE}.EITpatches)
```
agramfort: I would rename to : ${MODELBASE}-EIT.patches
unless .patches is different as the file format is different.

eolivi: a EIT patch has a surfacic extension. hence the 4th numeric column, this is the radius of the patch.
so do I rename it ?

OpenMEEG/libs/OpenMEEG/include/gain.h
```diff
@@ -213,13 +213,4 @@ namespace OpenMEEG {
         }
         ~GainStimInternalPot () {};
     };
-
-   class GainEITInternalPot : public Matrix {
-   public:
```
eolivi: as discussed with Theo, I should keep the GainEITInternalPot, and remove the GainStimInternalPot instead.
(this is not mentioned in the doc. only used once in openmeeg_sample_data)

papadop:
Merged. Thanks.

PR #174, opened 2017-03-29
Vtk continue

eolivi:
Build only the VTK libs that we need


OpenMEEG/tests/CMakeLists.txt
```diff
@@ -229,3 +229,9 @@ foreach(DIP 1 2 3 4 5 6)
         endforeach()
     endforeach()
 endforeach()
+
+if(WIN32)
+    STRING(REPLACE "/" "\\" CMAKE_MSVCIDE_RUN_PATH "${CMAKE_MSVCIDE_RUN_PATH}")
```
papadop: Use file(TO_NATIVE_PATH path result) instead ?

PR #181, opened 2017-04-07
setting deployment through scp on travis.

eolivi:


agramfort:
how do you make it work? with an ssh key?

agramfort:
it says in https://travis-ci.org/openmeeg/openmeeg/jobs/219690035 :

$ scp *gz $deploy_user@$deploy_host:$deploy_path
ssh_askpass: exec(/usr/X11R6/bin/ssh-askpass): No such file or directory
Host key verification failed.
lost connection

to make it work you might need to open a PR from a branch in https://github.com/openmeeg/openmeeg as private keys are not available on pull requests from forks.

PR #186, opened 2017-04-13
correct path for sftp from windows appveyor

eolivi:
from	
        build\OpenMEEG-2.3.0.2-Win64-OpenBLAS-static.tar.gz
to
	OpenMEEG-2.3.0.2-Win64-OpenBLAS-static.tar.gz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
         - USE_VTK=1
         - ENABLE_PACKAGING=1
         - BUILD_DOCUMENTATION=1
-       #- BLASLAPACK_IMPLEMENTATION=Auto (not specifying is the same as Auto)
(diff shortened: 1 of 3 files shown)


agramfort:
@eolivi are you also building packages for mac?

eolivi:
you are right.. it should, but it did not:
  File "/Users/travis/build/openmeeg/openmeeg/build_tools/upload_package_gforge.py", line 4, in <module>
    import paramiko
ImportError: No module named paramiko

one minute: I'll correct that here too.

BTW we only pack in tgz for Mac ? I think we will soon pack as well in DEB

agramfort:
yes +1 for .tgz on mac

for debian I think it's more tricky to do proper and have it accepted in
debian testing.

PR #189, opened 2017-04-24
[MRG] Restoring the possibility for EIT patches to have only 3 columns (not spatially extended).

eolivi:
after speaking with one of our user, we have figured out that before (release 2.1), it was possible to give EIT patch as a 3 columns matrix: x y z
whereas we have introduced later (at least in release2.3), the option for extended sensors as circular patches: x y z r
where r is the radius of the patch.
But It was badly done (probably me), and prohibited the use of EIT patches without spatial extension... 
this PR correct this behaviour. 

A test is added for a 3 column patch for Head2. but as before the gain computed is not compared to any analytical solution (since we don't have any for EIT).
[ anyway EIT remains validated through the specific test om_validate_EIT (test: EITvalidation-dipoles-Head) ]


eolivi:
appveyor did not run...

eolivi:
appveyor did not run....

eolivi:
appveyor did not run....

agramfort:
@eolivi let me know when CIs are happy

in the mean time I've the PR to [WIP]. Then make it [MRG]

eolivi:
I didn't get why you've put WIP. CIs are happy, and it waits for review

OpenMEEG/libs/OpenMEEG/src/sensors.cpp
```diff
@@ -225,7 +227,7 @@ namespace OpenMEEG {
     void Sensors::findInjectionTriangles() {
         om_error(m_geo!=NULL);
         m_weights = Vector(m_positions.nlin());
-        m_weights.set(0.);
+        m_weights.set(1.);
```
eolivi: no, each weight was overwritten ... but I've set the default case for the 3-columns EIT case

agramfort:
besides LGTM

it would be nice to keep track of all the bug fixes and changes.

see how we do with mne : http://martinos.org/mne/stable/whats_new.html#whats-new

this pages get's updated for every PR merged.

agramfort:
thx @eolivi

PR #191, opened 2017-05-05
Cgal surfacic

eolivi:
adding a tool to decimate properly a mesh, this tool keeps the topology of the input mesh, which is not always the case if you use the remesh tool.


eolivi:
Do not merge, because I want to find a better solution, because now, each cgal tool is about 80 MB, which is definitely too big. So I'll propbably do a intermediate library, or merge tools.

agramfort:
thanks @eolivi for looking into this yet I think this problem is solved already by MNE, Fieldtrip and Brainstorm. I don't think it's the role of OpenMEEG to take care of the geometry processing.

I also think that CGAL is absolutely not mandatory to make a release that is desperately needed. We can do it later.

my humble 2c...

eolivi:
yes, we have a user that was happy with our cgal tool, except for the decimation of the inner source space (lh.white.gii from freesurfer) where the topology was not preserved.
So Kai gave us his cgal tool for that. This is just the integration of such a tool.
p.s: does mne provide meshing tools ? only bem_watershed from freesurfer, nope ? our user did not want to use matlab

agramfort:
ok I see the background now. Yes mne has mne_watershed_bem and
mne_flash_bem if you have MRI Flash images.

OpenMEEG/libs/OpenMEEGMaths/include/matrix.h
```diff
@@ -188,8 +188,9 @@ namespace OpenMEEG {
     inline double Matrix::frobenius_norm() const {
     #ifdef HAVE_LAPACK
     if ( nlin()*ncol() != 0 ) {
-        double work;
-        return DLANGE('F',sizet_to_int(nlin()),sizet_to_int(ncol()),data(),sizet_to_int(nlin()),&work);
+        double work = -1.;
```
papadop: #define UNUSED(expr) do { (void)(expr); } while (0)
Add a call to this macro in BlasLapackImplementations/OpenMEEGMathsOpenBLASConfig.h

OpenMEEG/tools/CMakeLists.txt
```diff
@@ -38,26 +38,43 @@ add_executable(om_mesh_to_dip mesh_to_dip.cpp)
...
     endforeach()
     if (CGAL_ImageIO_FOUND) # these are tools hanling images (.inr format only for the moment!)
         foreach(target cgal_mesh_3Dlevelset_image cgal_mesh_3Dlabeled_image)
             add_executable(om_${target} ${target}.cpp)
             list(APPEND OM_CGAL_EXEC om_${target})
-            target_link_libraries(om_${target} OpenMEEG ${CGAL_LIBRARIES} ${VTK_LIBRARIES})
+            target_link_libraries(om_${target} OpenMEEG OpenMEEGCgal ${CGAL_LIBRARIES} ${VTK_LIBRARIES})
             target_include_directories(om_${target} PUBLIC ${CGAL_INCLUDE_DIRS})
             target_compile_options(om_${target} PUBLIC ${CGAL_CXX_FLAGS})
         endforeach()
```
papadop: Factorize ?

OpenMEEG/tools/cgal_decimate.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+The fact that you are presently reading this means that you have had
+knowledge of the CeCILL-B license and that you accept its terms.
+*/
+
+// Mesh simplification adapted from Kai Dang 2015 @ Inria
+
+#include <mesh.h>
+#include <options.h>
+#include "cgal_mesh.h"
```
papadop: Use <>

OpenMEEG/tools/cgal_mesh.cpp
```diff
@@ -0,0 +1,417 @@
+/*
```
papadop: Split in sevral files while keeping the idea of a library.

PR #192, opened 2017-05-12
Correct EIT RHS, and operatorP1P0

eolivi:
This solves a bug in operatorP1P0 present from eb008e50

### DIFF
--- a/OpenMEEG/libs/OpenMEEG/include/operators.h
+++ b/OpenMEEG/libs/OpenMEEG/include/operators.h
@@ -180,7 +180,10 @@ namespace OpenMEEG {
     }
 
     inline double _operatorP1P0(const Triangle& T2,const Vertex& V1) {
-        return (T2.contains(V1)) ? 0.0 : T2.area()/3.0;
+        double result = 0.;
+        if  (T2.contains(V1))
+            result = T2.area()/3.0;
+        return result;
     }
 
     template <typename T>
@@ -422,7 +425,7 @@ namespace OpenMEEG {
         // This time mat(i, j)+= ... the Matrix is incremented by the P1P0 operator
         std::cout << "OPERATOR P1P0... (arg : mesh " << m.name() << " )" << std::endl;
         for (Mesh::const_iterator tit = m.begin(); tit != m.end(); ++tit)
-            for (Mesh::VectPVertex::const_iterator pit = m.vertex_begin(); pit != m.vertex_end(); ++pit)
+            for (Triangle::const_iterator pit = tit->begin(); pit != tit->end(); ++pit)
                 mat(tit->index(), (*pit)->index()) += _operatorP1P0(*tit, **pit) * coeff;
     }
 
--- a/OpenMEEG/libs/OpenMEEG/src/assembleSourceMat.cpp
+++ b/OpenMEEG/libs/OpenMEEG/src/assembleSourceMat.cpp
@@ -142,6 +142,9 @@ namespace OpenMEEG {
     void assemble_EITSourceMat(Matrix& mat, const Geometry& geo, const Sensors& electrodes, const unsigned gauss_order)
     {
         //  A Matrix to be applied to the scalp-injected current to obtain the Source Term of the EIT foward problem.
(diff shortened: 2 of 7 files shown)


OpenMEEG/libs/OpenMEEG/include/operators.h
```diff
@@ -180,7 +180,10 @@ namespace OpenMEEG {
     }
 
     inline double _operatorP1P0(const Triangle& T2,const Vertex& V1) {
-        return (T2.contains(V1)) ? 0.0 : T2.area()/3.0;
+        double result = 0.;
+        if  (T2.contains(V1))
+            result = T2.area()/3.0;
+        return result;
     }
```
eolivi: there was the bug

OpenMEEG/libs/OpenMEEG/include/operators.h
```diff
@@ -422,7 +425,7 @@ namespace OpenMEEG {
         // This time mat(i, j)+= ... the Matrix is incremented by the P1P0 operator
         std::cout << "OPERATOR P1P0... (arg : mesh " << m.name() << " )" << std::endl;
         for (Mesh::const_iterator tit = m.begin(); tit != m.end(); ++tit)
-            for (Mesh::VectPVertex::const_iterator pit = m.vertex_begin(); pit != m.vertex_end(); ++pit)
+            for (Triangle::const_iterator pit = tit->begin(); pit != tit->end(); ++pit)
                 mat(tit->index(), (*pit)->index()) += _operatorP1P0(*tit, **pit) * coeff;
```
eolivi: not needed to iterate over all vertices each time ! only the vertex of the triangle

OpenMEEG/libs/OpenMEEG/src/sensors.cpp
```diff
@@ -153,10 +153,10 @@ namespace OpenMEEG {
         // weights
```
eolivi: just renaming radius -> radii

OpenMEEG/libs/OpenMEEG/src/assembleSourceMat.cpp
```diff
@@ -153,31 +156,35 @@ namespace OpenMEEG {
...
 
-        for (Geometry::const_iterator mit0=geo.begin();mit0!=geo.end();++mit0){
+        for (Geometry::const_iterator mit0 = geo.begin(); mit0 != geo.end(); ++mit0) {
             if (mit0->current_barrier()) {
-                for (Geometry::const_iterator mit1=geo.begin();mit1!=geo.end();++mit1){
-                    const int orientation=geo.oriented(*mit0,*mit1);
-                    if (orientation!=0){
-                        operatorS(*mit1,*mit0,transmat,geo.sigma_inv(*mit0,*mit1)*(-1.0*K*orientation),gauss_order);
-                        operatorD(*mit1,*mit0,transmat,(K*orientation),gauss_order,true);
+                for (Geometry::const_iterator mit1 = geo.begin(); mit1 != geo.end(); ++mit1) {
```
agramfort: we need to setup a linter for these style changes otherwise we'll keep arguing with @papadop 

personally I prefer with space. Let's just have robot to tell us what style to adopt.

agramfort:
LGTM

OpenMEEG/libs/OpenMEEG/include/operators.h
eolivi: of for you or shall I change that into:
return (T2.contains(V1))?T2.area()/3.0:0.;
as Theo suggest ?

eolivi:
P.S: the validateEIT test did not fail before nor now because the difference are at 10^-13 for this test.
This does not mean that some other cases were not impacted more.
Checking the Calderon Identities will soon cover the correctness of all our operators.

OpenMEEG/libs/OpenMEEG/include/operators.h
agramfort: this is way more readable for me. Please keep it like this.

eolivi:
merge ?

agramfort:
thanks @eolivi

PR #193, opened 2017-05-16
cosmit: changing for insensitive case option.

eolivi:


agramfort:
appveyor is not happy

LGTM otherwise

OpenMEEG/src/assemble.cpp
```diff
@@ -48,6 +48,10 @@ knowledge of the CeCILL-B license and that you accept its terms.
 #include <sensors.h>
 #include <geometry.h>
 
+#ifdef _MSC_VER
+#define strcasecmp _stricmp
+#endif
+
```
eolivi: I had to add that ...

agramfort:
+1 for merge

papadop:
Well, this is not equivalent to the existing code as now H2m and h2M will also be accepted... I'm not sure at all to like this change. This is against all practice (at least on unix like machines, I do not know for windows which might be case insensitive by principle in which case it'll be OK for windows) or MacOs, but I do not think that polluting one system by (poor) practices of another is good. I do not know why we have made such a complex setup initially.

papadop:
What does it try to solve ?

eolivi:
> this is not equivalent 

you are right.

As you want. it was just to save space, and not having to write all cases

agramfort:
> What does it try to solve ?

don't break API and make the code shorter and less bug prone?

eolivi:
I remove the useless "std::", and addedd the missing messages in somecases

OpenMEEG/src/assemble.cpp
```diff
@@ -50,27 +52,30 @@ knowledge of the CeCILL-B license and that you accept its terms.
...
 int main(int argc, char** argv)
 {
     print_version(argv[0]);
 
     bool OLD_ORDERING = false;
-    if ( argc<2) {
-        cerr << "Not enough arguments \nPlease try \"" << argv[0] << " -h\" or \"" << argv[0] << " --help \" \n" << endl;
+    if ( argc<2 ) {
+        getHelp(argv);
         return 0;
```
eolivi: instead of complaining: help the user. (do we go for that behavior in other binarys ?)

OpenMEEG/src/assemble.cpp
```diff
@@ -81,17 +86,11 @@ int main(int argc, char** argv)
...
-            std::cerr << "Please set conductivities filepath !" << endl;
-            exit(1);
-        }
+    if ( option(argv[1], {"-HeadMat","-HM", "-hm"}) ) {
         if ( argc < 5 ) {
-            std::cerr << "Please set output filepath !" << endl;
+            cerr << "Please set geometry filepath" << endl;
+            cerr << "           conductivities filepath" << endl;
+            cerr << "           output filepath" << endl;
             exit(1);
```
papadop: A better message is to say that option HeadMat requires 3 files (geometry,conductivity and output) and only XX have been provided.

papadop: I propose that this message is integrated to the option function by passing the list of files (then the number can be decided and the message generated/adapted. And even much less lines of code !!!!

agramfort: agreed. It's good practice to say what was wrong instead of just saying how it should be.

agramfort:
@papadop merge when you're happy

papadop:
There is still a minor glitch because a , is at the end of the list of parameters.

PR #194, opened 2017-05-19
Packaging on Windows

eolivi:
As we have changed our ways of handling dlls:
Before: copy all dll to the bin folder
Present: do not copy any Dll to bin, only specify the PATH where the dlls are.

So what was missing was that: the dll s were no more in the package.

Now from the SP packaging we call all subprojects packagings


cmake/Packaging.cmake
```diff
@@ -15,10 +15,14 @@ if (ENABLE_PACKAGING)
     set(CPACK_GENERATOR "TGZ")
 
     # if we want to generate all the sub-project packages:
-    #set(CPACK_INSTALL_CMAKE_PROJECTS "${CMAKE_CURRENT_BINARY_DIR}/matio/build;matio;ALL;/" "${CMAKE_CURRENT_BINARY_DIR}/OpenMEEG/build;OpenMEEG;ALL;/")
-
-    # call cpack for OpenMEEG
     set(CPACK_INSTALL_CMAKE_PROJECTS "${CMAKE_CURRENT_BINARY_DIR}/OpenMEEG/build;OpenMEEG;ALL;/")
+    foreach (dep ${MSINTTYPES} zlib ${VTK_} hdf5 matio ${LAPACK})
```
papadop: I'm merging. I just wonder whether it would not be smart to introduce an OpenMEEG_DEPENDENCIES (or sthg like that) and use it everywhere. Otherwise there is the chance that we change the list in the main CMakeList.txt and not here.....

papadop:
See above....

cmake/Packaging.cmake
eolivi: yes you're right

PR #196, opened 2017-05-29
when building static: link against static matio and openblas for Open…

eolivi:
…MEEGMaths.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ matrix:
         - USE_CGAL=1
         - USE_GIFTI=1
         - USE_VTK=1
-        - BUILD_SHARED_LIBS=OFF
+        - BUILD_SHARED_LIBS=ON
         - BLASLAPACK_IMPLEMENTATION=Atlas
 
 before_install: source build_tools/travis/before_install.sh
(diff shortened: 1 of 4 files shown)


agramfort:
can you do the same for hdf5?

it works here for me so far.

eolivi:
I though this would work... but for hdf5, we would need to specify all the libs as static*:
HDF5_LIBRARIES shared are: /usr/lib64/libhdf5_cpp.so;/usr/lib64/libhdf5.so;/usr/lib64/libz.so;/usr/lib64/libdl.so;/usr/lib64/libm.so 

and it is not doable on linux, since the libs are not installable as static libs (no apt-get install hdf5-static matio-static ,... ).

it would work on windows (since we compile ourself the zlib, hdf5, matio,) ... but that's not the problem.

1. I'll go for the complete static build fixing the issue with IOs ?
 or 
2. I do a hack, and set HDF5_LIBRARIES to libhdf5_cpp.a libhdf5.a ?

agramfort:
what I had done in the past was to set a default file extension based on BUILD_SHARED_LIBS so it was looking for all .a files for static build. I did not have to change it everywhere.

agramfort:
see how it was done in release 2.2

https://github.com/openmeeg/openmeeg/blob/release-2.2/macros/UseAtlas.cmake#L14

at that time we only had the problem with Blas/Lapack

you need to change the order in the CMAKE_FIND_LIBRARY_SUFFIXES variable

eolivi:
closing this as redundant with https://github.com/openmeeg/openmeeg/pull/197

PR #197, opened 2017-05-29
when building static: link against static libs if you can for OpenMEE…

eolivi:
…GMaths.


papadop:
OK for me except the removal of the white line in AsciiIO.H ;-)

eolivi:
this will be ok for MAC but does not work on linux (anyway... not needed)

agramfort:
can I test the produced binaries if they work?

eolivi:
yes you can. it should be ok now.
I did on our MAC from ci:
ciosx:bin ci$ otool -l om_assemble  | grep -i hdf
ciosx:bin ci$ otool -l om_assemble  | grep -i mat

and no output (and it used to (with the shared matio and hdf5)).

agramfort:
it is not uploaded to http://openmeeg.gforge.inria.fr/download/?C=M;O=D ?

eolivi:
no it will upload only from the master....
so you have to test from my branch called static2

agramfort:
we should be able to upload an openblas version static from travis right?

agramfort:
I am merging so I can test the mac binaries

agramfort:
@eolivi I downloaded http://openmeeg.gforge.inria.fr/download/OpenMEEG-2.3.0.2-MacOSX-Intel-OpenBLAS-static-vtk.tar.gz and it works on my machine but I needs openblas

```
$ ldd om_assemble 
om_assemble:
	/usr/local/opt/openblas/lib/libopenblasp-r0.2.19.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	libOpenMEEGMaths.1.dylib (compatibility version 1.0.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
```

it should be possible to use libopenblas.a cf.

$ ls /usr/local/opt/openblas/lib/libopenblas.a 
/usr/local/opt/openblas/lib/libopenblas.a

also this was built by ci.inria.fr or travis? if ci.inria.fr I thought we were using it to use MKL Inria license?

eolivi:
done in https://github.com/openmeeg/openmeeg/pull/198

PR #198, opened 2017-05-31
adding Openblas.a for static builds

eolivi:


eolivi:
... does not seem to be possible, on my mac:
[  1%] Linking CXX shared library libOpenMEEGMaths.dylib
Undefined symbols for architecture x86_64:
  "__gfortran_concat_string", referenced from:
      _dormbr_ in libopenblas.a(dormbr.o)
      _dormlq_ in libopenblas.a(dormlq.o)
      _dormqr_ in libopenblas.a(dormqr.o)
  "__gfortran_pow_i4_i4", referenced from:
      _dlasd0_ in libopenblas.a(dlasd0.o)
      _dlasda_ in libopenblas.a(dlasda.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

eolivi:
I would go for the other solution which is to build fully in static mode (without having to link against the static libs of matio, hdf5, openblas,...).
This would make the static build available on all platforms, at the expense of a header to include in all the executables .cpp (to load the IOs).
What do @papadop  says ? and @agramfort  ?

agramfort:
any idea why it's asking me to find clapack lib when building with OpenBLAS static on my mac?

agramfort:
>
> What do @papadop <https://github.com/papadop> says ? and @agramfort
> <https://github.com/agramfort> ?
>

whatever works !

make it work then make it nice :)

papadop:
The initial problem as I see it, is that libopenblas.a requires the addition of a libgfortran (.so or .a).
Static libraries, contrarily to dynamic ones (and that's maybe no longer totally true) do not record the libraries they are depending on...
All the remaining stuff in OpenMEEGMaths is totally unrelated (and thus unneeded cruft)... I do not see how it can solve this specific problem.

eolivi:
@agramfort 
what if we do as for windows, we deliver the package WITH the libs openblas, matio, hdf5 ?
-r--r--r--  1 ci  admin   2.9M Feb 15 09:48 /usr/local/opt/hdf5/lib/libhdf5.100.dylib
-r--r--r--  1 ci  admin   214K Feb 15 09:49 /usr/local/opt/libmatio/lib/libmatio.4.dylib
-r--r--r--  1 ci  admin    67M Feb 15 15:49 /usr/local/opt/openblas/lib/libopenblasp-r0.2.18.dylib

ok for you ? (Theo is ok)

agramfort:
whatever works. If François Tadel tells me the brainstorm users have a working system I am happy. That's my only figure of merit.

PR #199, opened 2017-05-31
when building static: fully build static libs, and  load IOs.

eolivi:


OpenMEEG/src/gain.cpp
```diff
@@ -78,7 +78,6 @@ int main(int argc, char **argv)
             cerr << "Not enough arguments \nPlease try \"" << argv[0] << " -h\" or \"" << argv[0] << " --help \" \n" << endl;
             return 0;
         }
-        LinOpInfo matinfo = OpenMEEG::maths::info(argv[3]);
         SymMatrix HeadMatInv;
```
eolivi: what was this line for ?

build_tools/travis/before_install.sh
```diff
@@ -24,7 +24,7 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
         brew link openblas --force  # required as link is not automatic
     fi
 
-    if [[ "$USE_VTK" == "1" && $BUILD_SHARED_LIBS != OFF ]]; then
+    if [[ "$USE_VTK" == "1" ]]; then
         brew install vtk
```
eolivi: no more need for building our own static vtk. (faster builds on travis)

cmake/Packaging.cmake
```diff
@@ -15,7 +15,7 @@ if (ENABLE_PACKAGING)
     set(CPACK_GENERATOR "TGZ")
 
     # if we want to generate all the sub-project packages:
-    set(CPACK_INSTALL_CMAKE_PROJECTS "${CMAKE_CURRENT_BINARY_DIR}/OpenMEEG/build;OpenMEEG;ALL;/")
+    set(CPACK_INSTALL_CMAKE_PROJECTS)
```
eolivi: just because OpenMEEG was packed twice

PR #201, opened 2017-06-02
adding tag 2.4.prerelease (and package version 2.3.99)

eolivi:


eolivi:
is it ok for you to specify 2.3.99 as a pre-release version ?
seen there: https://cmake.org/pipermail/cmake/2009-May/029450.html

agramfort:
what @maedoc means is to do something like

git tag v2.3.2
git push --tags

eolivi:
yes I know, I can do that once the version as changed internally (= this PR accepted)

agramfort:
ok for me

PR #206, opened 2017-06-08
Default options

agramfort:
I was annoyed to have to manually disable doxygen and python.

I don't think these should default to ON.

I also went over the install instructions.

Feedback welcome.


.travis.yml
```diff
@@ -22,20 +22,22 @@ matrix:
         - ENABLE_PACKAGING=1
         - BUILD_SHARED_LIBS=ON
         - USE_VTK=1
-        - BUILD_DOCUMENTATION=1
+        - BUILD_DOCUMENTATION=OFF
```
agramfort: let's only build the doc once on linux and once on mac

agramfort:
@papadop @eolivi ok for you?

I am done here

.travis.yml
eolivi: you also have to change into before_install.sh:
>     if [[ "$BUILD_DOCUMENTATION" == "1" ]]; then
        brew install Doxygen Graphviz # For building documentation

because else travis is not happy

eolivi:
beside my comment: ok for me.

agramfort:
@eolivi green !

agramfort:
thx @eolivi

PR #207, opened 2017-06-08
Optim

eolivi:


README.rst
```diff
@@ -76,7 +76,7 @@ On Mac OS X, you'll need `CMake <http://www.cmake.org>`_ (install it with `Homeb
 
     *optionally*::
 
-    $ brew install hdf5 && brew install libmatio --with-hdf5
+    $ brew install hdf5 && brew install libmatio
```
agramfort: please revert. it's in my other PR :)

eolivi: yes I've seen.
I'll wait for other comments ?

agramfort:
nope that's it for me

agramfort:
appveyor is not happy

eolivi:
Making OperatorD Optimzed and parallel is a big optimization as it reduces the time for HM-Head3 from 45s to 18s

build_tools/travis/before_install.sh
```diff
@@ -28,6 +28,10 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
         brew install vtk
     fi
 
+    if [[ "$USE_OMP" == 1 ]]; then
+        brew install llvm
+    fi
+
```
eolivi: @agramfort  please is that the correct formula we need to build with openMP ? or clang support it by default ?

agramfort: I have no idea.

Before clang we had gcc 4.2 on mac and I did the previous release using the intel compiler anyway...

agramfort:
so you removed the non-optimized D operator?

OpenMEEG/libs/OpenMEEG/include/operators.h
```diff
@@ -394,7 +315,13 @@ namespace OpenMEEG {
         //
 
         unsigned i = 0; // for the PROGRESSBAR
-        for (Mesh::const_iterator tit1=m1.begin();tit1!=m1.end();++tit1) {
+        #pragma omp parallel for
+        #ifndef OPENMP_3_0
+        for (int i1=0; i1 < m1.size(); ++i1) {
+            const Mesh::const_iterator tit1 = m1.begin()+i1;
```
agramfort: is this the improvement?

eolivi: yes !

eolivi: yes

build_tools/travis/before_install.sh
```diff
@@ -28,6 +28,10 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
         brew install vtk
     fi
 
+    if [[ "$USE_OMP" == 1 ]]; then
+        brew install llvm
```
agramfort: ok I tested. It does the job but you need to change the compiler that cmake is using.

for me I need to set:

CPP = /usr/local/opt/llvm/bin/clang
CPPFLAGS = -I/usr/local/opt/llvm/include -fopenmp
LDFLAGS = -L/usr/local/opt/llvm/lib

this being said we should not ship binaries with clang-omp / llvm on mac as it will break on most systems that don't brew setup properly.

We should ship non-openmp binaries on mac and update the brew formula to add an openmp option.

So to sum up:
- setup/update only one travis build so it uses openmp using brew setting up properly all env vars.
- upload to gforge non-openmp binaries for our users

eolivi: >CMake Error at /usr/local/Cellar/cmake/3.8.2/share/cmake/Modules/CMakeTestCCompiler.cmake:51 (message):
  The C compiler
  "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang"
  is not able to compile a simple test program.
  It fails with the following output:
   Change Dir: /Users/travis/build/openmeeg/openmeeg/build/OpenMEEG/build/CMakeFiles/CMakeTmp
  /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang
  -fopenmp=libomp -w -o CMakeFiles/cmTC_f4998.dir/testCCompiler.c.o -c
  /Users/travis/build/openmeeg/openmeeg/build/OpenMEEG/build/CMakeFiles/CMakeTmp/testCCompiler.c
  clang: error: unsupported argument 'libomp' to option 'fopenmp='

for now I have that, while I've set:
        export CXX="/usr/local/opt/llvm/bin/clang"
        export CPPFLAGS="$CPPFLAGS -I/usr/local/opt/llvm/include -fopenmp"
        export LDFLAGS="$LDFLAGS -L/usr/local/opt/llvm/lib"

I'm trying on our ci mac machine... but it takes ages to brew llvm

agramfort:
travis passes with this?

build_tools/travis/before_install.sh
agramfort: you're still using xcode clang and not the one from brew. See

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang

in travis log.

you need to set flags for C and not only CPP files.

eolivi:
arf... cannot try this on my Mac from ci, since I cannot install llvm (at least with brew)...
It says that MacOsX 10.9 is too old for this...

agramfort:
ok let try

agramfort:
I tried adding

        export CC="/usr/local/opt/llvm/bin/clang"
        export CXX="/usr/local/opt/llvm/bin/clang"
        export CPPFLAGS="$CPPFLAGS -I/usr/local/opt/llvm/include -fopenmp"
        export CFLAGS="$CFLAGS -I/usr/local/opt/llvm/include -fopenmp"
        export LDFLAGS="$LDFLAGS -L/usr/local/opt/llvm/lib"

but I get a link error. Can you try with travis?

agramfort:
travis is not happy at all here.

this PR is the top priority.

I would leave out OMP for mac to converge quickly.

I have still hope to have a prelease before OHBM

eolivi:
I think it is okay now. with OpenMP.
The error was
> export CXX="/usr/local/opt/llvm/bin/clang"

which should be
> export CXX="/usr/local/opt/llvm/bin/clang++"

eolivi:
good to review/merge . it works with OpenMP
previously:
HM-HeadNNc2 ...................................   Passed    8.25 sec
whereas now:
HM-HeadNNc2 ...................................   Passed    3.86 sec

agramfort:
great !

thanks @eolivi 

do we have binaries to test on windows and linux now?

PR #210, opened 2017-06-17
add cpplint to travis

agramfort:
@papadop @eolivi thoughts?

you'll see that there are hundreds of style issues

I am not saying cpplint is the gold standard but let's just agree on "one" standard we all agree to respect.


agramfort:
you see all errors in https://travis-ci.org/openmeeg/openmeeg/jobs/243975580

but it does not make travis crash. This can certainly be fixed.

@eolivi please take over. I'll stop on this.

agramfort:
go go @eolivi !

agramfort:
@eolivi does travis fail if there is a linting error?

agramfort:
@papadop can you live with this?

eolivi:
no it does not fail hopefully !
still about 600 errors, whereas 3000 at first (ok I added filters too )
I think we should merge that soon as a good first step toward cpplint compliancy (because it is possible without conflicts now),
and soon go toward uncrustify + cpplint

agramfort:
good to go on my end.

@papadop you don't get mad if I press the green button ?

papadop:
I may get mad.... Give me some time to check what this cpplint does... We have waited for so long tha it's not a matter of a week.... At first glance, cpplint seems very restrictive (google coding rules and very little customization).

eolivi:
regarding this PR, I mostly only added/removed spaces (as suggested by cpplint).
I agree it may not be exactly what we want, but this is already a good step forward.
Have a look please to the files or the commit names to see what I actually changed.
I would like this PR merged soon (<= 1 week), because else it will be hard to merge later (many git conflicts to solve).

agramfort:
I agree that we should agree ASAP on a convention and then stick to it.
No need to twist our brains too much.

Here is an uncrustify config that matches cpplint conventions:

https://gist.githubusercontent.com/philectron/1a1e1ccb06d5643cd61d85fc5ecaf429/raw/8813567b8f7dd49003228e19c3d10719b49995a1/uncrustify-preferences.cfg

see
https://stackoverflow.com/questions/43244788/uncrustify-google-style-c-config

let's keep the ball rolling !

eolivi:
@papadop please tell when you're ok for merging at least this PR (as a first step toward beautiful c++), and I will resolve the conflicts with the master.
Then I'll do a PR with uncrustify , etc...

massich:
@agramfort can you rebase?

agramfort:
closing in favor of #270 

but we may need to argue with @papadop about the guidelines we want to follow / impose.

PR #212, opened 2017-06-21
Update tutorial version and date

eolivi:


agramfort:
LGTM

it would be great to convert this .tex file to rst and insert them in the
main website:

https://github.com/openmeeg/openmeeg.github.io

you can put maths in rst as in:

http://scikit-learn.org/stable/modules/linear_model.html#lasso

eolivi:
so no more latex ? LGTMT*

I can remove the latex dependence from OpenMEEG then ?


* = LGTM Too :-D

agramfort:
yes no more Latex !

eolivi:
so the rst will be converted into html (with docutils ?) ? because how can we have colors in rst ? 
https://stackoverflow.com/questions/4669689/how-to-use-color-in-text-with-restructured-text-rst2html-py-or-how-to-insert-h

agramfort:
sphinx does all the heavy work for you.

this is what I use to build the doc.

need my help?

PR #215, opened 2017-06-24
tell cmake to use rpath to avoid having to set DYLD_LIBRARY_PATH

agramfort:
this avoids the need to set DYLD_LIBRARY_PATH on my mac after doing a make install in a custom folder.

maybe this can be helpful for our problem on mac @eolivi

### DIFF
--- a/OpenMEEG/cmake/OpenMEEGOptions.cmake
+++ b/OpenMEEG/cmake/OpenMEEGOptions.cmake
@@ -1,3 +1,7 @@
+# Tell CMake to use rpath with the libs we build
+set(CMAKE_MACOSX_RPATH 1)
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+
 # Set CMAKE_BUILD_TYPE to Release by default.
 if (NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


OpenMEEG/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake ${CMAKE_SOURCE_DIR}/cmake/Fortra
 
 project(OpenMEEG LANGUAGES C CXX ${PROJECT_VERSION})
 
+# Tell CMake to use rpath with the libs we build
+set(CMAKE_MACOSX_RPATH 1)
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+
 # The next two lines are a kludge to make rpm generation work. If you change this,
 # please check that rpm generation still works.
```
eolivi: you should put thes 3 lines in OpenMEEGOption.cmake, it gets included in both CMakeLists.txt (through Setup.cmake)

eolivi:
Please put the label MergeMe when you're ready also ;-)

agramfort:
please take over the PR. I have to task switch

PR #211, opened 2017-06-19
add an option so that not all libs (vtk, cgal,...) are always copied …

eolivi:
…on Mac

PR #217, opened 2017-06-26
onMac: test whether the package is correct.

eolivi:


build_tools/travis/after_success.sh
```diff
@@ -28,3 +28,15 @@ if [[ $ENABLE_PACKAGING == "1" && $TRAVIS_PULL_REQUEST == "false" && $TRAVIS_BRA
     conda install -y --quiet pyopenssl
     python ${src_dir}/build_tools/upload_package_gforge.py *gz *deb
 fi
+
+# now test the binary packages
+if [[ "$ENABLE_PACKAGING" == "1" && $APPLE_STANDALONE == "1" ]]; then
+    # uninstall all
+    brew uninstall --ignore-dependencies --force hdf5 libmatio vtk cgal openblas
```
agramfort: use

/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall)"

see

https://github.com/Homebrew/install

eolivi: it will remove all installed formulas ?

agramfort:
yes it will remove all brew

eolivi:
Dear @agramfort ,
The package test fails because:
> dyld: Library not loaded: /usr/local/opt/szip/lib/libsz.2.dylib
  Referenced from: /Users/travis/build/openmeeg/openmeeg/build/OpenMEEG-2.3.99.0-MacOSX-Intel-OpenBLAS-standalone-cgal/./bin/om_assemble
  Reason: image not found


following our discussion there https://github.com/openmeeg/openmeeg/pull/211

I definitely think copying all libs is NOT DOABLE* and IS NOT THE RIGHT SOLUTION for a standalone package, so I give up on this task.
Either we go for non standalone (Mac users either install OM through brew or have the dependencies), or we go for Bundle package manager.

*: I've tried to copy/make symlink in the lib install dir, but although it worked for libz, it didnot for libsz...

agramfort:
to install a bundle on mac you need to be admin as far as I know.
Brainstorm mac users get OpenMEEG without noticing because it
just takes the .tar.gz and expand it.

while it has always worked with a .tar.gz using static linking on mac
let's try to find a hack to make it work.

I see 2 options:
- we don't use brew for the standalone (we use matio+hdf5 from
superproject) and vecLib from the system. It will be slower than openmp +
openblas but at least it will work out of the box.


- since we have travis to make the binaries you can: untar the zip,
add the libs I told you manually to the lib and then tar again.

wdyt?

eolivi:
done in https://github.com/openmeeg/openmeeg/pull/218

PR #218, opened 2017-06-27
Solves issue94, correct few lines after static analysis, move patches…

eolivi:
… folder to cmake/external


agramfort:
can you add a test for the new check conductivity ?

besides LGTM

agramfort:
fine with me if CIs are happy as well as @papadop

OpenMEEG/libs/OpenMEEG/include/OMstd.h
```diff
@@ -0,0 +1,60 @@
+/*
+Project Name : OpenMEEG
```
eolivi: is the name OMstd.h ok for you ?

OpenMEEG/libs/OpenMEEG/src/sensors.cpp
```diff
@@ -207,7 +207,7 @@ namespace OpenMEEG {
             if (hasOrientations())
                 outfile << m_orientations.getlin(i) << " ";
             // if it has weights (other than 1)
-            if (std::abs(m_weights.sum() - m_weights.size()) > 1.e3*std::numeric_limits<double>::epsilon()) {
+            if (almost_equal(m_weights.sum(), static_cast<double>(m_weights.size()))) {
                 outfile << m_weights(i) << std::endl;
```
eolivi: bug here ! add "not"

agramfort:
@eolivi appveyor is not happy but travis seems to have succeeded building the standalone version. How can I check?

eolivi:
yes it looks good !
I've added a prefix ("PR-")to the package so that it is uploaded to the forge even when coming from a PR.

./bin/om_assemble [-option] [filepaths...]
option :
   -HeadMat, -HM, -hm :   
       Compute Head Matrix for Symmetric BEM (left-hand side of linear system).
             Arguments :
               geometry file (.geom)
               conductivity file (.cond)
...

agramfort:
where is the .tar.gz ?

eolivi:
> I've added a prefix ("PR-")to the package so that it is uploaded to the forge even when coming from a PR.

actually the upload doesn't work, because the password variable are only available on master.

So let us review and merge ?

agramfort:
alright let's see

PR #219, opened 2017-06-28
Properly add exception in properties

papadop:
Properties.H was lax in the way it was handling things. Tidied up. It is now the work of the caller to do the proper messages and exit.


agramfort:
I'll let you guys sync with #218

PR #223, opened 2017-07-10
Gifti

eolivi:
Add gifti/nifti as a subproject so that installation is automatic


agramfort:
is this a python PR or Gifti PR?

forcing python 2.7 in 2017 seems like a bad idea.

is there an urgent need for gifti support?

eolivi:
people here always ask me gifti and complain that they have to install it  (pfff PhDs ... ;-) )
because the meshes from freesurfer use that format.

By default the python wrapping was not working (because targetting python 3), so as we provide binding for 2.7 I've just added this.

But we have planned with Theo this afternoon to check 3.0 wrapping or/and fixing the bug

agramfort:
>
> people here always ask me gifti and complain that they have to install it
> (pfff PhDs ... ;-) )
>
tell them to ask on github next time :)

> because the meshes from freesurfer use that format.
>
freesurfer does not use gifti but it's own surf file.

I wrote the IO for freesurfer meshes in nibabel.

there are both python and matlab code for such IO.

we already have a lot of code in OpenMEEG which has nothing to do with BEM
solvers (IO, our own linalg classes etc.). Let's avoid too much technical
debts with stuff that are not central to what OpenMEEG is good for.

> By default the python wrapping was not working (because targetting python
> 3), so as we provide binding for 2.7 I've just added this.
>
ok then let's make a dedicated Python PR and don't mix this with gifti

> But we have planned with Theo this afternoon to check 3.0 wrapping or/and
> fixing the bug
>

ok great. Now as we discussed before Python is not mandatory for the
release but just "nice to have".
We should focus on the release.

FYI : François Tadel told me the windows binaries work on brainstorm for
him in windows 10

eolivi:
close for now, as gifti / nifti cmake is such in a bad condition .

PR #224, opened 2017-07-10
remove universal build for mac (only intel for years now)

agramfort:
it's been > 10years there is only intel mac now.


agramfort:
ok for you @papadop @eolivi ?

agramfort:
thx

PR #225, opened 2017-07-11
Make the build python3 ready.

papadop:
Not the correction of the infamous python bug... But an improvement for those who might build openmeeg on a machine where python is the default.


eolivi:
what about adding in this PR the python test on travis and appveyor then (+the notification of the caveat regarding memory and temporary objects ) ?
(I can do it if you want (If I have write access on your repo )).

agramfort:
+1 to check CIs build before merging this.

thx @papadop for looking into this.

OpenMEEG/wrapping/src/test_python.py
```diff
@@ -1,4 +1,20 @@
...
+# > M = om.SymMatrix(100)
+# taking as an numpy array the sub-matrix of M might lead to corrupted memory:
+# > mySubMat = om.asarray(M.submat(0,10,0,10))
+# since submat returns a newly created object that is hold by nothing, thus
+# destroyed afterward.
+# Instead do keep an object pointing the newly created submatrix, and then
+# access the numpy array form of it:
+# > subM = M.submat(0,10,0,10)
+# > mySubMat = om.asarray(subM)
+###########################################
```
agramfort: this message should be put in:

https://raw.githubusercontent.com/openmeeg/openmeeg_sample_data/master/compute_leadfields.py

thanks a lot

eolivi: yes ok.
beside that .. we have a failing test on mac (test_python) and travis says green (because we added the check package for sure...)
99% tests passed, 1 tests failed out of 1165

Total Test time (real) =  85.50 sec

The following tests FAILED:

	348 - openmeeg_python (OTHER_FAULT)

agramfort:
you use python from brew with mac so if you remove brew you don't get an ok package.

I would recommend to use python with anaconda on mac as it's the most standard install.

but honestly I think this can be done after the release... no urgent need to release Python binaries.

OpenMEEG/wrapping/src/test_python.py
eolivi: corrected (setting set -e in script.sh)

eolivi:
I've given up testing python on either MAC or ubuntu.
Always config problems with numpy ...

agramfort:
fair enough. Let's leave python out for now

eolivi:
can I merge ?

OpenMEEG/wrapping/src/CMakeLists.txt
```diff
@@ -13,51 +11,46 @@ include_directories(${PYTHON_INCLUDE_PATH} ${NUMPY_INCLUDE_DIRS} ${CMAKE_CURRENT
 find_package(SWIG REQUIRED)
 include(${SWIG_USE_FILE})
 
-set(MODULE_NAME openmeeg)
-set(SWIG_SRC_FILE_NAME ${MODULE_NAME}.i)
```
agramfort: by removing this you made the diff quite big. Was this necessary?

agramfort:
do we test on both python2 and python3 ?

eolivi:
no, we test only on windows with python27.
I gave up about travis as said before.

But anyway I for merging that

agramfort:
you had issues with python on linux and mac? even using conda?

eolivi:
yes

agramfort:
@papadop do you want this in the release?

I'd rather have full python support for all 3 systems tested before
releasing this.

papadop:
We already ship (only in source mode) the python biding. This is a net progress and I do not think there can be regressions because of this work. So I'm in favor of merging instead of letting it rot in some branch. If there was a regression (including only in testing), merging this work could be questionable, but in this case, it has explicitly the same testing as before (we just allow people with python3 the possibility of using the binding and testing it).

If there is no other counter argument, I'll merge later tonight or tomorrow....

build_tools/travis/before_install.sh
```diff
@@ -87,7 +87,7 @@ else
     fi
 
     if [[ "$BUILD_DOCUMENTATION" == "ON" ]]; then
-        sudo apt-get install doxygen graphviz
+        sudo apt-get install -y doxygen graphviz gnuplot
```
agramfort: ok to merge but can you just remove gnuplot here? I don't see why this is needed. thx

agramfort:
closing in favor of #234

this branch adds travis builds for python 2.7 and python 3.6 using conda

PR #226, opened 2017-07-14
on cleaning cmake targets

eolivi:
@papadop  please, you told me you could do something concerning the targets.

The problems are: 
- that they do not depend correctly on each other.
- the default target 'ALL' need to depend on build (that should itself depend on build-OpenMEEG AND install-OpenMEEG, ... )
- there are targets created by ExternalProject function (in cmake/SetTargets.cmake), they are called OpenMEEG-configure (or more generraly ${ep}-${target}), whereas the one you use in EP_AddCustomTargets are called build-OpenMEEG, ... etc (or more generraly ${target}-${ep}). I think this one source of the problems.

I've remove one custom clean target as it is dupplicated (cmake says), but please check dependencies.

Thank you

### DIFF
--- a/cmake/externals/ConfigureExternalProjects.cmake
+++ b/cmake/externals/ConfigureExternalProjects.cmake
@@ -111,6 +111,7 @@ macro(subprojects)
         && echo To build them, run the build target: 'cmake --build . --target build'
     )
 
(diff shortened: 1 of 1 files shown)


massich:
I think we can close this one @agramfort @papadop @eolivi

PR #230, opened 2017-07-22
remove brew and use curl no wget for conda

agramfort:


eolivi:
regarding your PR ! good (i just did not know how to use curl instead of wget.)

regarding vtk on mac Lapack. I forgot something. I can fix that.

I've just checked the linux versions on a single CPU VirtualMachine.
So for the Lapack version: building Head3 took: 71 s and inversion : 53 s
Using the Openblas package: took 68 s and I could not invert (because missing lapack.so)

So I would say, let us forget openblas for linux-package and go for lapack ? 
Openblas maybe faster when parallel is on but, it will be the same for inversion since they did not optimized dsptrf which is why we need liblapack.so as well (see https://github.com/xianyi/OpenBLAS/issues/182#issuecomment-12461750:
> OpenBLAS implements optimized variants of a subset of the LAPACK functions. 
)


and the Openblas - MAC version ?

agramfort:
see : https://travis-ci.org/openmeeg/openmeeg/jobs/256353154

to see the packaging pb on mac.

ok for our own lapack on linux if this works.

for mac if you make packaging pass travis test I am happy with any solution.
Yes openblas would be nice.

thx

PR #232, opened 2017-07-24
packaging: switch back to vecLib standalone on mac and clapack on linux.

eolivi:
finally switch back to previous standalone with vecLib on Mac,
and for linux, simply build delivering everything matio, zlib, hdf5 and clapack as time is comparable to openblas (comparison done on a 4-cpu VM: assembly of headMat3 took around 80s in all cases, and similar inversion).

This way the code is simpler too (no ugly copying libs and sym links)


eolivi:
@papadop @agramfort 
if the inverse is so slow compared to mkl, we can still deliver with MKL either:
- using ci.inria.fr
- to build one package home made and upload it to the forge (this would be faster than 1st option, but it won't be continuous dev)

agramfort:
openblas is not tested anymore on mac

see https://github.com/travis-ci/travis-ci/issues/4604

for MKL on linux

apparently you can apt-get install now

eolivi:
> openblas is not tested anymore on mac

right.. DONE

> apparently you can apt-get install now

whene ? no you cannot and you wont get the mkl-dev with the headers we need

agramfort:
>
> right.. DONE
>
great

> apparently you can apt-get install now
>
> whene ? no you cannot and you wont get the mkl-dev with the headers we need
>

arfff if you say so.

FYI I just had the confirmation that the mac openblas binaries at:

http://openmeeg.gforge.inria.fr/download/OpenMEEG-2.3.99.0-MacOSX-OpenBLAS-standalone-vtk-cgal.tar.gz

where working. But if you say it's not faster...

PR #233, opened 2017-07-25
improving gainEEG and gainMEG on memory footprints.

eolivi:
is it ok like this Theo @papadop ?
I did not check the memory was actually better like this, but it should.

When calling Matrix(const &tmp) it does not the DEEP_COPY right ?

I've also removed the useless (??) lines  :
        LinOpInfo matinfo = OpenMEEG::maths::info(argv[3]);

### DIFF
--- a/OpenMEEG/libs/OpenMEEG/include/gain.h
+++ b/OpenMEEG/libs/OpenMEEG/include/gain.h
@@ -53,6 +53,7 @@ namespace OpenMEEG {
     class GainMEG : public Matrix {
     public:
         using Matrix::operator=;
+        GainMEG (const Matrix& GainMat): Matrix(GainMat) {}
         GainMEG (const SymMatrix& HeadMatInv,const Matrix& SourceMat, const Matrix& Head2MEGMat, const Matrix& Source2MEGMat) {
             *this = Source2MEGMat+(Head2MEGMat*HeadMatInv)*SourceMat;
         }
@@ -62,6 +63,7 @@ namespace OpenMEEG {
     class GainEEG : public Matrix {
     public:
         using Matrix::operator=;
+        GainEEG (const Matrix& GainMat): Matrix(GainMat) {}
         GainEEG (const SymMatrix& HeadMatInv,const Matrix& SourceMat, const SparseMatrix& Head2EEGMat) {
             *this = (Head2EEGMat*HeadMatInv)*SourceMat;
         }
--- a/OpenMEEG/libs/OpenMEEG/include/gain.h
+++ b/OpenMEEG/libs/OpenMEEG/include/gain.h
@@ -53,21 +53,31 @@ namespace OpenMEEG {
     class GainMEG : public Matrix {
     public:
         using Matrix::operator=;
-        GainMEG (const Matrix& GainMat): Matrix(GainMat) {}
-        GainMEG (const SymMatrix& HeadMatInv,const Matrix& SourceMat, const Matrix& Head2MEGMat, const Matrix& Source2MEGMat) {
+
+        GainMEG(const char * HeadMatInv, const char * SourceMat, const char * Head2MEGMat, const char * Source2MEGMat) {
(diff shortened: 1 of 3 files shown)


agramfort:
2 concerns here
now it seems that the old constructor of
GainMEG
GainEEG
with all the matrices is useless.
also we have 2 ways of doing the same thing: pass char* to constructor or use load. We should have only one way to do things. If we keep the constructor with char* let's remove the load method.

also the constructor as in `GainMEG MEGGainMat(tmp);` that takes a matrix as input without making a copy seems dangerous/misleading.

my 2c

eolivi:
yes it it more or less useless now to use the classes GainEEG and GianMEG.
That was a short fix so that it is included in the release.

I won't do better now on this.

OpenMEEG/libs/OpenMEEG/include/gain.h
```diff
@@ -53,6 +53,7 @@ namespace OpenMEEG {
     class GainMEG : public Matrix {
     public:
         using Matrix::operator=;
+        GainMEG (const Matrix& GainMat): Matrix(GainMat) {}
         GainMEG (const SymMatrix& HeadMatInv,const Matrix& SourceMat, const Matrix& Head2MEGMat, const Matrix& Source2MEGMat) {
             *this = Source2MEGMat+(Head2MEGMat*HeadMatInv)*SourceMat;
```
agramfort: what you did if more efficient than? I think the GainEEG and GainMEG classes were introduced for the Python bindings. If you fix the constructor here it will also benefit Python users right?

OpenMEEG/src/gain.cpp
```diff
@@ -113,17 +107,9 @@ int main(int argc, char **argv)
...
-        Matrix SourceMat;
-        SourceMat.load(argv[3]);
-        Matrix Head2MEGMat;
-        Head2MEGMat.load(argv[4]);
-        Matrix Source2MEGMat;
-        Source2MEGMat.load(argv[5]);
+        const Matrix &tmp = Matrix(argv[5]) + (Matrix(argv[4]) * SymMatrix(argv[2])) * Matrix(argv[3]);
 
-        GainMEG MEGGainMat(HeadMatInv, SourceMat, Head2MEGMat, Source2MEGMat);
+        GainMEG MEGGainMat(tmp);
```
agramfort: @papadop I don't like much this new API. Now it's not explicit anymore what MEGGainMat does. 

Also I don't get why this is more efficient than doing:

*this = Source2MEGMat+(Head2MEGMat*HeadMatInv)*SourceMat;

as it is done in the constructor so far.

OpenMEEG/libs/OpenMEEG/include/gain.h
eolivi: yes it was for the python bindings ... So it is still possible to use it for python. 
But if the user is short on memory he should do himself the gain operations releasing each matrix ASAP when doing the products....
I don't see another way for doing that... or maybe with filenames instead of matrices for the constructor.... ?

OpenMEEG/src/gain.cpp
eolivi: because each matrix is being released after the computation of the product or addition

OpenMEEG/libs/OpenMEEG/include/gain.h
eolivi: ok I'm doing it with filenames... better indeed

eolivi:
> I've also removed the useless (??) lines :
LinOpInfo matinfo = OpenMEEG::maths::info(argv[3]);


ok I've done filename based constructors for gains... please review

agramfort:
Nice. I like it. Thanks @eolivi 

@papadop ok for you?

Let's merge after CIs are happy

eolivi:
and @papadop
what was these lines for ?
 LinOpInfo matinfo = OpenMEEG::maths::info(argv[3]);

papadop:
See PR #240 instead....

PR #234, opened 2017-07-28
[MRG] Python wrapping

agramfort:
i am trying to take over #225 to make it work with all 3 platforms


agramfort:
@papadop good to go from my end. Please merge if you're happy. I took your commits and just added travis tests for python 2.7 and 3.6

papadop:
Neat.... Thank's for making the testers work with python. A minor nitpick, why use wget on linux and curl on mac. curl is also available in linux. We can homogeneize there. Will do that in a followup commit on day...

PR #235, opened 2017-07-29
Fix linux packaging

agramfort:
linux binaries are still broken as vtk libs are missing from package.

I am setting up travis to show the bug and I will remove vtk from distrib until it's fixed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,14 +21,14 @@ matrix:
     #     - USE_PYTHON=1
     #     - PYTHON_VERSION=3.6
     #     - APPLE_STANDALONE=1
-    - os: linux
-      compiler: clang
-      env:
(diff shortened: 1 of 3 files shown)


agramfort:
ok so vtk was not fully removed from system and the failing linux package was therefore not detected. See:

https://travis-ci.org/openmeeg/openmeeg/builds/258861317?utm_source=github_status&utm_medium=notification

I am removing now vtk from the release on Linux unless @eolivi or @papadop make it work.

agramfort:
ok all green. merging

agramfort:
FYI these binaries at:

http://openmeeg.gforge.inria.fr/download/OpenMEEG-2.3.99.0-Linux-LAPACK-OpenMP.tar.gz

work on my server at Télécom.

PR #238, opened 2017-07-30
add Gifti automatic support

eolivi:
but not automatically tested on travis since sometimes the sourceforge which download nifti/gifti stalls....


agramfort:
I am against merging anything that is not tested. If needs be make a minor on github. It must exists already I am sure.

eolivi:
as you want.... it worked on my machine but problems on the old trusty because it is linking -libm and it is not the correct way (I would have to patch again the nifti... )

agramfort:
Honestly I don't really see the point of this PR

We have already a code base too big for the engineering time we have and we keep adding dependencies. We should focus on quality. Cf bug reported on brainstorm forum yesterday with new binaries. FYI I am presently on holidays and we have the deadline for the BACI conf for a stable release.

larsoner:
I too am inclined to support *less* mesh I/O support inside OpenMEEG if possible and let people use mesh conversion routines instead. Okay to close @eolivi ?

PR #240, opened 2017-07-30
Memory

papadop:
Supercedes pull request #233.
Merge of eolivi's version and papadop one.
+ Added some tools to follow memory allocations (not activated by default).

### DIFF
--- a/OpenMEEG/src/gain.cpp
+++ b/OpenMEEG/src/gain.cpp
@@ -161,11 +161,14 @@ int main(int argc, char **argv)
             return 0;
         }
         const SymMatrix HeadMatInv(argv[2]);
-        const Matrix SourceMat(argv[3]);
         const Matrix Head2IPMat(argv[4]);
+
+        const Matrix& tmp1 = Head2IPMat*HeadMatInv;
+        const Matrix SourceMat(argv[3]);
+        const Matrix& tmp2 = tmp1*SourceMat;
         const Matrix Source2IPMat(argv[5]);
 
-        const GainInternalPot InternalPotGainMat(HeadMatInv, SourceMat, Head2IPMat, Source2IPMat);
+        const Matrix& InternalPotGainMat = Source2IPMat+tmp2;
         InternalPotGainMat.save(argv[6]);
     }
     else if ( (!strcmp(argv[1], "-EITInternalPotential"))||(!strcmp(argv[1], "-EITIP")) ) {
@@ -174,10 +177,13 @@ int main(int argc, char **argv)
             return 0;
         }
         const SymMatrix HeadMatInv(argv[2]);
-        const Matrix SourceMat(argv[3]);
         const Matrix Head2IPMat(argv[4]);
 
-        const GainEITInternalPot InternalPotGainMat(HeadMatInv, SourceMat, Head2IPMat);
+        const Matrix& tmp = Head2IPMat*HeadMatInv;
+
+        const Matrix SourceMat(argv[3]);
+
+        const Matrix& InternalPotGainMat = tmp*SourceMat;
         InternalPotGainMat.save(argv[5]);
     }
     else {
--- a/OpenMEEG/src/gain.cpp
+++ b/OpenMEEG/src/gain.cpp
@@ -78,16 +78,14 @@ int main(int argc, char **argv)
             cerr << "Not enough arguments \nPlease try \"" << argv[0] << " -h\" or \"" << argv[0] << " --help \" \n" << endl;
             return 0;
         }
-        //const SymMatrix HeadMatInv(argv[2]);
-        //const SparseMatrix Head2EEGMat(argv[4]);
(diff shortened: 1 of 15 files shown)


OpenMEEG/src/gain.cpp
```diff
@@ -132,18 +127,14 @@ int main(int argc, char **argv)
...
-        Matrix Source2MEGMat;
-        Source2MEGMat.load(argv[7]);
-
-        GainMEGadjoint MEGGainMat(geo, dipoles, HeadMat, Head2MEGMat, Source2MEGMat);
+        const Matrix dipoles(argv[4]);
+        const SymMatrix HeadMat(argv[5]);
+        const Matrix Head2MEGMat(argv[6]);
+        const Matrix Source2MEGMat(argv[7]);
+
+        const GainMEGadjoint MEGGainMat(geo, dipoles, HeadMat, Head2MEGMat, Source2MEGMat);
```
agramfort: why using GainMEGadjoint here? can't you spare memory here too?

agramfort: forget it

OpenMEEG/src/gain.cpp
```diff
@@ -174,35 +160,27 @@ int main(int argc, char **argv)
...
-        Matrix Source2IPMat;
-        Source2IPMat.load(argv[5]);
-
-        GainInternalPot InternalPotGainMat(HeadMatInv, SourceMat, Head2IPMat, Source2IPMat);
+        const SymMatrix HeadMatInv(argv[2]);
+        const Matrix SourceMat(argv[3]);
+        const Matrix Head2IPMat(argv[4]);
+        const Matrix Source2IPMat(argv[5]);
+
+        const GainInternalPot InternalPotGainMat(HeadMatInv, SourceMat, Head2IPMat, Source2IPMat);
```
agramfort: I think that here you can spare memory too

OpenMEEG/src/gain.cpp
```diff
@@ -174,35 +160,27 @@ int main(int argc, char **argv)
...
-        SourceMat.load(argv[3]);
-        Matrix Head2IPMat;
-        Head2IPMat.load(argv[4]);
-
-        GainEITInternalPot InternalPotGainMat(HeadMatInv, SourceMat, Head2IPMat);
+        const SymMatrix HeadMatInv(argv[2]);
+        const Matrix SourceMat(argv[3]);
+        const Matrix Head2IPMat(argv[4]);
+
+        const GainEITInternalPot InternalPotGainMat(HeadMatInv, SourceMat, Head2IPMat);
```
agramfort: here too

agramfort:
my only concern is that GainMEG / GainEEG etc. are not tested anymore right?

papadop:
To answer here. I have closed #233 before seeing the last comments. So answering here...
The linop is unnecessary as far as I can tell, I already removed it in my PR.
I'm not too keen on having the GainEEG stuff working with file names.... I prefer that all IOs are made in the main cpp not in libraries.
Ideally, as GainEEG and consort are inlined the compiler might be able to optimize things by himself but I suspect that because of IOs it is not abl to do it.... Maybe there is a way to help it, but I do not see how to do this for now.... If I have tie I may ask why this is not optimized on the gcc mailist....
As for the testing of the GainEEG and consort, they still should get tested by the python code (no ?). This is unverified though. I  now see them more as helpers for API users than anything else.

papadop:
For the other optimizations, yes. Originally I only looked at GainEEG, GainMEG was in eolivi's  version. I can have a look at the other ones later tonight.

agramfort:
I also think that filenames in constructor is fine as it leaves the logic of the code in the same place ie. we have some code duplication but it's just below so clearly visible.

yes Python executes the GainEEG and GainMEG but it does not test the correctness of the results AFAIK. It would be nice to do though.

papadop:
OK I added the other improvements indicated by Alex.
I'd expect people using python to do some matrix manipulations, so I do not see them using the "filename" form of the GainXXX classes... I'd also prefer to use the GainXXX classes in gain.cpp, but for now we cannot probably only because of compiler limitations.... I'd expect the compiler to do the optimization using the original formulations (after all all the matrices are marked as const so no aliasing may occur and the compiler should be allowed to reorder the calls and free the memory incrementally. I need to understand why this is not the case. I'm not totally opposed to have the loading constructor in the GainXXX classes, but this will be alternate implementations on the normal constructors and I really do not see people using python saving the previous matrices in files and then calling these classes. They will probably have the matrices already loaded and will use either the formula directly or the GainXX helper classes with their original constructors....

agramfort:
@papadop you have my commits in your branch. Can you rebase on the current master to have a clean git history? thx

papadop:
Sorry. I have been trying for over than an hour to get this clean, but did not manage...  I prefer to merge now. In this way, Il will see what is a non-clean history ;-) ....

papadop:
Merged...

PR #244, opened 2017-10-12
C++ 11

papadop:
New fedora version = new compiler and new python3 => several errors, which are corrected with this patch...


agramfort:
travis is not happy

papadop:
Yes but not my fault I think.... It seems to be that the clang version usés to compile python (fromage conda) and thé one installes and used to compile opemreg are not compatible... I'm not sure we should block for that as anyway we do not ship versions compiled with clang....Le 12 oct. 2017 8:32 PM, Alexandre Gramfort <notifications@github.com> a écrit :travis is not happy

agramfort:
travis is green so unless you prove me that travis has changed and current
master fails on new travis I don't want to merge code that breaks our
continuous integration.

papadop:
OK merging this PR, and will redo the other one.

PR #245, opened 2017-10-12
H2 ecogm

papadop:
Implement the old behaviour for h2ecogm. The choice is made on the number of parameters. I chosed to not conditionalize this  on the matrix version, even though the version number is now available for assemble.cpp.... but this is to be discussed.

The new innermost_interface detects the innermost interface (!) which is 2 for Alexamdre's example and not 3 as in the test. With this, it give the same matrices.

I think that there is also a subtle bug that I corrected in the detection of nested geometries (the dit_out variable was wrongly set. Maybe this is related to the TODO comment:
 // TODO test model HeadNNa0 appears as non-nested even if it is nested
This needs to be checked....

Tests still need to be added... I can do it if someone provides the proper files. Alex example is not that big (only the dipole file is, but we do not use it)... Use it ?

### DIFF
--- a/OpenMEEG/libs/OpenMEEG/include/assemble.h
+++ b/OpenMEEG/libs/OpenMEEG/include/assemble.h
@@ -51,85 +51,85 @@ namespace OpenMEEG {
 
     class OPENMEEG_EXPORT HeadMat: public virtual SymMatrix {
     public:
-        HeadMat (const Geometry& geo, const unsigned gauss_order=3);
+        HeadMat (const Geometry& geo,const unsigned gauss_order=3);
         virtual ~HeadMat () {};
     };
 
     class OPENMEEG_EXPORT SurfSourceMat: public virtual Matrix {
     public:
-        SurfSourceMat (const Geometry& geo, Mesh& sources, const unsigned gauss_order=3);
+        SurfSourceMat (const Geometry& geo,Mesh& sources,const unsigned gauss_order=3);
         virtual ~SurfSourceMat () {};
     };
 
     class OPENMEEG_EXPORT DipSourceMat: public virtual Matrix {
     public:
-        DipSourceMat (const Geometry& geo, const Matrix& dipoles, const unsigned gauss_order=3,
-                      const bool adapt_rhs = true, const std::string& domain_name = "");
+        DipSourceMat (const Geometry& geo,const Matrix& dipoles,const unsigned gauss_order=3,
+                      const bool adapt_rhs = true,const std::string& domain_name = "");
         virtual ~DipSourceMat () {};
     };
 
     class OPENMEEG_EXPORT EITSourceMat: public virtual Matrix {
     public:
-        EITSourceMat(const Geometry& geo, const Sensors& electrodes, const unsigned gauss_order=3);
+        EITSourceMat(const Geometry& geo,const Sensors& electrodes,const unsigned gauss_order=3);
         virtual ~EITSourceMat () {};
     };
 
     class OPENMEEG_EXPORT Surf2VolMat: public virtual Matrix {
     public:
         using Matrix::operator=;
-        Surf2VolMat(const Geometry& geo, const Matrix& points);
+        Surf2VolMat(const Geometry& geo,const Matrix& points);
         virtual ~Surf2VolMat () {};
     };
 
     class OPENMEEG_EXPORT Head2EEGMat: public virtual SparseMatrix {
     public:
-        Head2EEGMat (const Geometry& geo, const Sensors& electrodes);
+        Head2EEGMat (const Geometry& geo,const Sensors& electrodes);
         virtual ~Head2EEGMat () {};
     };
 
     class OPENMEEG_EXPORT Head2ECoGMat: public virtual SparseMatrix {
     public:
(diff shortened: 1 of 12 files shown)


OpenMEEG/libs/OpenMEEG/include/assemble.h
```diff
@@ -51,85 +51,85 @@ namespace OpenMEEG {
...
+                const unsigned gauss_order=3,double alpha=-1.,double beta=-1.,const std::string &filename="");
         virtual ~CorticalMat () {};
     };
 
     class OPENMEEG_EXPORT CorticalMat2: public virtual Matrix {
     public:
-        CorticalMat2(const Geometry& geo, const Head2EEGMat& M, const std::string& domain_name = "CORTEX",
-                const unsigned gauss_order=3, double gamma=1., const std::string &filename="");
+        CorticalMat2(const Geometry& geo,const Head2EEGMat& M,const std::string& domain_name = "CORTEX",
+                     const unsigned gauss_order=3,double gamma=1.,const std::string &filename="");
```
agramfort: @papadop we really need a style checker on travis as otherwise I'll keep undoing your styles changes. I put spaces after , ...

OpenMEEG/libs/OpenMEEG/src/assembleSensors.cpp
```diff
@@ -45,142 +45,145 @@ knowledge of the CeCILL-B license and that you accept its terms.
...
-            for ( unsigned k = 0; k < 3; ++k) {
-                current_position(k) = positions(it, k);
-            }
-            dist_point_interface(current_position, i, current_alphas, current_triangle);
-            mat(it, current_triangle.s1().index()) = current_alphas(0);
-            mat(it, current_triangle.s2().index()) = current_alphas(1);
-            mat(it, current_triangle.s3().index()) = current_alphas(2);
+        for (unsigned it = 0; it<positions.nlin(); ++it) {
+            for (unsigned k = 0; k<3; ++k)
+                current_position(k) = positions(it,k);
```
agramfort: same here.

please please @papadop put cosmit changes in dedicated PRs as you make my review harder with many green lines that are totally unrelated to the fix

agramfort:
also you added the other commits to this branch.

please tell me where to look.

for the test I would just write manually a text file for sEEG and ECoG on a sphere model. Center is 0 and we know the radii.

papadop:
For the style I agree.... But doing a specific cosmetic changes is complicated. I usually do the changes while correcting other bugs. Usually it starts just around the place where I need to do changes, and after a while it percolates... I do changes while reading the code (and it helps me understanding the code), I will not redo the work twice (once to introduce the cosmetic changes and once for the actual changes). I tried to summarize the important changes in the initial message...
Introduction of a VersionId enum and member  in the Geometry class (which is unused now, and I not too decided on what to do) and use it in the geometry reader (here introducing the enum is a net reading imnprovement, so at least I would keep the enum in that class). In the reader there is also the dit_out change in the nested model detection. The remainder is in assemble.cpp with the duplication of the H2ECOGM option and the introduction (in geometry.{h,cpp}) of the innermost_interface function.

The C++-11 changes (well strictly speaking the python ones) were needed to be able to compile....

agramfort:
I won't ask you to revert the cosmits but it is not complicated to stick to a code style if it's well accepted and in the continuous integration. I asked @eolivi to do it to specifically avoid the situation of this PR.

ok for C+11 fixes but then close the other PR.

also travis is not happy on one build.

thanks a lot @papadop for taking a stab at this. Let's make CIs green and add tests and we are good to go.

OpenMEEG/src/assemble.cpp
```diff
@@ -268,25 +268,49 @@ int main(int argc, char** argv)
...
+
+        // Assembling Matrix from discretization:
+        // Head2ECoG is the linear application which maps x |----> v
+
+        std::cerr << "Warning: we assume that ECoG electrodes are placed on the inner interface." << std::endl
+                  << "This is only valid for nested files. Consider specifying an interface as a name" << std::endl
+                  << " right after the electrode position file." << std::endl;
+        const Interface& innerSkullLayer = geo.innermost_interface();
+        std::cerr << "Detected interface: " << innerSkullLayer.name() << ' ' << geo.interface("3").name() << std::endl;
+        Head2ECoGMat mat(geo,electrodes,innerSkullLayer);
```
eolivi: why is it std::cerr and not cout ?
why is it the interface "3" ? interfaces can be named in the geom file.

eolivi:
beside my comment, this LGTM even if it is hard to review as Alex said and should better be re-submitted after the c++11 PR passes, and cosmit are submitted aside and so on... 
(you still have some lines with spaces | tabs only).

BTW if you have  modified a lot of file with both cosmit and functional changes, you can separate the commits with this git command:
` git commit -p ` it will promt you what you want to commit interactively.

papadop:
The steer message was a test I forgot to remove...
The "3" here was referring to the example Alex gave me.  Comparison tests were not working because the inner layer was "2" and not "3". Anyhow this has to be suppressed and yes the names "1" and "2" are bad but are those of the example.

papadop:
Thanks for the commit -p. I may try to do the separation to exercise....

papadop:
I will point out the other bug I think I have corrected.  You might confirm it....

PR #246, opened 2017-10-14
[MRG] Fixing mac CI build

agramfort:


agramfort:
ok this should fix the mac failing build. merging.

PR #247, opened 2017-10-15
Update OpenMEEG version to 2.4

eolivi:
the tools were still showing they were compiled with 2.3.99
please accept this PR as soon as you think we're really good for release 2.4


agramfort:
thx @eolivi

PR #248, opened 2017-10-18
New H2ECOGM branch

papadop:
Here is a new "clean" PR... I hope I forgot nothing.... And git commit -p was totally un-useful. Did it by hand... I'm not abandonning the other changes, I'll just push them in another PR to be cleaner....
Let's see what Travis says, and I will add Maureen tests before merging.


OpenMEEG/src/assemble.cpp
```diff
@@ -268,25 +268,51 @@ int main(int argc, char** argv)
...
+        // Reading the file containing the positions of the EEG patches
 
-        // read the file containing the positions of the EEG patches
         Sensors electrodes(argv[4]);
 
-        // Assembling Matrix from discretization :
+        // Assemble matrix from discretization:
+        // Head2ECoG is the linear application which maps x |----> v
+
+        std::cerr << "Warning: we assume that ECoG electrodes are placed on the inner interface." << std::endl
```
agramfort:
```
(WARNING) New version of H2ECOGM option requires you to specify the name of the interface on which ECoG positions are positioned. Consider specifying an interface as a name
 right after the electrode position file. Without this information, we will consider that the ECoG electrodes are placed on the inner interface.
```
besides I read when running this PR with the demo ECoG model:

Detected interface: 2 3

I don't know what this means.

agramfort:
I confirm that the logic works and that I can still use old files now.

agramfort:
also this file: https://github.com/openmeeg/openmeeg/blob/master/OpenMEEG/data/README.rst

should be migrated to tutorial and pdf. So far it's hard to find and the command line points me to it. It should point to the tutorial page.

OpenMEEG/src/assemble.cpp
papadop: It was a debug message I used to debug the example you gave me. It was not giving the same result. The interface 2 is the one detected as the inner one and "3" is the one that was in the script. Removed with the new commit.

papadop:
This can be done in a separate PR. I'll add tomorrow the test cases Maureen prepared, This will improve the coverage. Then I will merge this branch if everything goes well (beside the Mac travis pb). Let's do the documentation change elsewhere....

agramfort:
Fair enough

Alex

papadop:
Not totally perfect as the epsilons used in the ECoG tests are very big due to the way we compute errors, but overall it works.

OpenMEEG/src/assemble.cpp
```diff
@@ -401,7 +411,11 @@ int main(int argc, char** argv)
 bool option(const int argc, char ** argv, const Strings& options, const Strings& files) {
     for ( auto s: options) {
         if (argv[1] == s) {
-            if (argc-2 < files.size()) {
+            unsigned minimum_nparms = files.size();
```
agramfort: are you testing this option somewhere?

papadop: Yes. I made or to merge the two versions of H2ECoGM treating the index pf the surface as optional. This is what allowed me to merge the two.  There are other option with optional parameters that can now be documented in the option function. Previously, they were not because not providing one would have generated a wrong number of parameters error.

papadop: Yes. I made or to merge the two versions of H2ECoGM treating the index pf the surface as optional. This is what allowed me to merge the two.  There are other option with optional parameters that can now be documented in the option function. Previously, they were not because not providing one would have generated a wrong number of parameters error.

papadop:
On 10/18/2017 06:07 PM, Alexandre Gramfort wrote:
> also this file:
> https://github.com/openmeeg/openmeeg/blob/master/OpenMEEG/data/README.rst
> 
> should be migrated to tutorial and pdf. So far it's hard to find and the
> command line points me to it. It should point to the tutorial page.

Interestingly, it seems that mkl is now totally free !!!
The script above does nor work easily because it targets intel compilers
so it requires a serial key and that generates some diffculties...

But we hacked a pure cmake version with Maureen....
A pull request will appear shortly...

	Theo.

PR #253, opened 2017-11-14
Mkl on travis and appveyor

papadop:
Looking at MKL licence, it seems we can build with MKL as it is now free to use.
This PR extends FindMKL to download and install mkl when it is not found and MKL is required.
This allows travis and appeveyor builds hopefully.
Tested on linux fedora. Expect some troubles on windows as I'm not sure even that the installer is OK (the web site gives me the same linux/windows/apple link). We will see what appveyor tells us.


papadop:
Static build does not work with 2017.4.196. Works with 2018.0.128 but not with 2018.1.163 !!!

agramfort:
don't bother too much with windows. Our binaries with openblas do already
work.

how do I test on mac? if is set BLASLAPACK_IMPLEMENTATION to MKL it will
download
and install automatically? if so where does it install? do I have to be
root?

papadop:
I do bother of windows... despite the accelerations we did, it is now slower !!!
> how do I test on mac? if is set BLASLAPACK_IMPLEMENTATION to MKL it will
> download and install automatically? 
Yes.

> if so where does it install? do I have to be root?
In a subdirectory of the build directory. No need to be root. That being said, this is not totally convenient for a regular developper as mkl does strange things. For example, it records somewhere that there was an installation and keep thinking this even if you rm -rf the directory. This is probably not too problematic on travis/appveyor as we get a new machine everytime... We can restrict this to some CI-MODE if we find it too problematic....

Also, I'm almost sure that the downloaded package is wrong for MacOS (the URL does not seems to change when I change the arch, so if you provide me with the proper URL - I expect it will be correct if you get it from a MAC - I will correct that). Take the version 2018.0 (not 2017.4 or 2018.1 which seem to be buggy).

massich:
It seems I can't build up the PR on my system either

papadop:
Can you share a trace of the build (by email). Also, this is somewaht expected if your system is a Mac or a windows.... It should download the mkl package (for linux!) and then it will fail.... As I said, I need to update the URLs for Mac and windows. If this is on linux, the trace will be interesting (what linux distrib and version ?).

papadop:
As far as I can tell, the installation works on travis (all mkl files are there). It is the use in FindMKL which fails (trying to see if this is because we pass a cmake list instead of a string). Maybe be this is related to a cmake version, as it works on my computer with cmake 3.9....

papadop:
I finally found the culprit (I think). OpenMEEG MKL support requires the use of mkl_link_tool to define the proper flags and libraries to use (depending on the options). But mkl_link_tool is a 32 bits application, so it was failing with a "No such file or directory" even while the file was there.... This was really puzzling. I had to install a ubuntu virtualbox to find that.... I suspect travis will pass now... Let's see in a few minutes....

papadop:
Finally it worked  !!! Everything built and tests passed. I do not know why travis reports the build as a failure... Maybe the last lines of the log as they indicate trying to use om_assemble ?!?
Now, I really need for the mkl URLs for Mac and Windows....

massich:
[CMakeOutput.log](https://github.com/openmeeg/openmeeg/files/1478393/CMakeOutput.log)

agramfort:
see the linux error on travis:

https://travis-ci.org/openmeeg/openmeeg/jobs/302966731

it's because it's linking dynamically with vtk it says on testing at the
end:

./bin/om_assemble: error while loading shared libraries: libvtkIO.so.5.8:
cannot open shared object file: No such file or directory

I would disable vtk for now to test

massich:
I drop, the MWE i was doing. I'll try to put an example in travis using ICC

papadop:
Yeah. I was not seeing the last line earlier. libvtk is removed just a few lines above, so it should be very easy to correct. I suppose this happens in the after

agramfort:
the idea of removing vtk before testing the binaries is to guarantee that
binaries are self contained

agramfort:
@papadop only this failed: https://travis-ci.org/openmeeg/openmeeg/jobs/306183557

MKL not found on the linux build. The mac builds are green :)

agramfort:
still MKL not found 

https://travis-ci.org/openmeeg/openmeeg/jobs/307089965

we're close !

agramfort:
ok MKL is found but you seem to link to dynamic libs : 

https://travis-ci.org/openmeeg/openmeeg/jobs/308383114

papadop:
Yes. This is needed as otherwise I get a test error !!!
Overall, I believe shared libraries are better (but we should distribute them). Packages are smaller, and we do not get hidden multiple link problems...
But when I tried that on MacOS, it creates troubles (do not remember which ones). Each platform works with a different set of rules. This is a nightmare !!!

papadop:
On the positive side. The build now mostly succeeds for windows but we are hit by the one hour limit as vtk is also built (a trimmed version). I may try to ask for a larger limit, because I do not see what else to do. There seems to be no binary packaging of the full C++ development VTK kit...
Definitely, windows and Mac (to a lower degree) are not made for developers....

papadop:
Actually one build almost completed on appveyor.... Tests are just failing (probably because MKL dynamic libraries are not found)....

massich:
## Observations

As far as I got, the problem seems to be this one: The **mkl** libraries does not get linked statically, and the dynamic library is not on the path. 

```sh
(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ ./bin/om_assemble 
./bin/om_assemble: error while loading shared libraries: libmkl_intel_lp64.so: cannot open shared object file: No such file or directory
```

This can bee inspected with **ldd, readelf** and **ldd**. By looking at **ldd**, the shared libraries are not found. 

```sh
(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install/bin$ ldd om_assemble | grep libmkl
        libmkl_intel_lp64.so => not found
        libmkl_gnu_thread.so => not found
        libmkl_core.so => not found
        libmkl_intel_lp64.so => not found
        libmkl_gnu_thread.so => not found
        libmkl_core.so => not found
        libmkl_intel_lp64.so => not found
        libmkl_gnu_thread.so => not found
        libmkl_core.so => not found
```

Adding mkl to `LD_LIBRARY_PATH` should help us somehow.

```sh
(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ export LD_LIBRARY_PATH="/opt/intel/compilers_and_libraries_2018.0.128/linux/mkl/lib/intel64_lin/:$LD_LIBRARY_PATH"                                                
```

However it does not. Libraries still don't get found. By looking at the `rpath` we can see where the libraries are searched. (My guess is that when `rpath` is defined the libraries are only searched for in such path.

```sh
(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ objdump -x ./bin/om_assemble | grep -i path
  RPATH                /home/travis/openmeeg/openmeeg/build/OpenMEEG/install/lib



(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ readelf -d ./bin/om_assemble 

Dynamic section at offset 0x15d48 contains 37 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libOpenMEEG.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libOpenMEEGMaths.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_intel_lp64.so]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_gnu_thread.so]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_core.so]
 0x0000000000000001 (NEEDED)             Shared library: [libgomp.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libz.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000f (RPATH)              Library rpath: [/home/travis/openmeeg/openmeeg/build/OpenMEEG/install/lib]
 0x000000000000000c (INIT)               0x4052e8
 0x000000000000000d (FINI)               0x41121c
 0x0000000000000019 (INIT_ARRAY)         0x615d20
 0x000000000000001b (INIT_ARRAYSZ)       24 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x615d38
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x400298
 0x0000000000000005 (STRTAB)             0x402258
 0x0000000000000006 (SYMTAB)             0x400818
 0x000000000000000a (STRSZ)              8874 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000003 (PLTGOT)             0x616000
 0x0000000000000002 (PLTRELSZ)           2376 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x4049a0
 0x0000000000000007 (RELA)               0x404808
 0x0000000000000008 (RELASZ)             408 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffe (VERNEED)            0x404738
 0x000000006fffffff (VERNEEDNUM)         4
 0x000000006ffffff0 (VERSYM)             0x404502
 0x0000000000000000 (NULL)               0x0
```

So adding the libraries at the proper place should work.
```sh
(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ for ii in mkl_intel_lp64.so libmkl_gnu_thread.so libmkl_core.so; do
> ln -s $(find /opt/intel/ |  grep intel64 | grep $ii) ./lib/;
> done

(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ ln -s /home/travis/openmeeg/openmeeg/build/hdf5/install/lib/libhdf5.so.1.8.14 ./lib/libhdf5.so.9.0.0


(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ ls -l lib
total 2528
lrwxrwxrwx 1 travis travis      71 Nov 29 16:41 libhdf5.so.9.0.0 -> /home/travis/openmeeg/openmeeg/build/hdf5/install/lib/libhdf5.so.1.8.14
lrwxrwxrwx 1 travis travis      86 Nov 29 17:07 libmkl_core.so -> /opt/intel/compilers_and_libraries_2018.0.128/linux/mkl/lib/intel64_lin/libmkl_core.so
lrwxrwxrwx 1 travis travis      92 Nov 29 17:07 libmkl_gnu_thread.so -> /opt/intel/compilers_and_libraries_2018.0.128/linux/mkl/lib/intel64_lin/libmkl_gnu_thread.so
lrwxrwxrwx 1 travis travis      92 Nov 29 17:07 libmkl_intel_lp64.so -> /opt/intel/compilers_and_libraries_2018.0.128/linux/mkl/lib/intel64_lin/libmkl_intel_lp64.so
lrwxrwxrwx 1 travis travis      21 Nov 29 13:14 libOpenMEEGMaths.so -> libOpenMEEGMaths.so.1
lrwxrwxrwx 1 travis travis      25 Nov 29 13:14 libOpenMEEGMaths.so.1 -> libOpenMEEGMaths.so.1.1.0
-rw-r--r-- 1 travis travis  651062 Nov 29 13:12 libOpenMEEGMaths.so.1.1.0
lrwxrwxrwx 1 travis travis      16 Nov 29 13:14 libOpenMEEG.so -> libOpenMEEG.so.1
lrwxrwxrwx 1 travis travis      20 Nov 29 13:14 libOpenMEEG.so.1 -> libOpenMEEG.so.1.1.0
-rw-r--r-- 1 travis travis 1916468 Nov 29 13:12 libOpenMEEG.so.1.1.0
drwxrwxr-x 3 travis travis    4096 Nov 29 13:14 python3.6
```

So as expected now it passes

```sh
(wrappingenv) travis@7dab43e7bcf2:~/openmeeg/openmeeg/build/OpenMEEG/install$ ./bin/om_assemble 
./bin/om_assemble version 2.3.99 compiled at Nov 29 2017 13:12:51

./bin/om_assemble [-option] [filepaths...]

option :
   -HeadMat, -HM, -hm :   
       Compute Head Matrix for Symmetric BEM (left-hand side of linear system).
             Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               output matrix

   -CorticalMat, -CM, -cm :   
       Compute Cortical Matrix for Symmetric BEM (left-hand side of linear system).
       Comment on optional parameters:
       Giving two (or zero) numeric optional parameters => CorticalMat will try to use (or estimate) alpha/beta.
       Giving one numeric optional parameters => CorticalMat2 will use gamma.
       Giving a filename (a string), one can save time saving the intermediate matrix in all cases (useful when trying multiple values).
             Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               file containing the positions of EEG electrodes (.patches)
               domain name (containing the sources)
               output matrix
               [optional parameter alpha or gamma or filename]
               [optional parameter beta or filename]
               [optional filename]

   -SurfSourceMat, -SSM, -ssm :   
       Compute Surfacic Source Matrix for Symmetric BEM (right-hand side of linear system). 
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               mesh of sources (.tri .vtk .mesh .bnd)
               output matrix

   -DipSourceMat, -DSM, -dsm:    
      Compute Dipolar Source Matrix for Symmetric BEM (right-hand side of linear system). 
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               dipoles positions and orientations
               output matrix
               (Optional) domain name where lie all dipoles.

   -EITSourceMat, -EITSM -EITsm : 
       Compute the EIT Source Matrix from an injected current (right-hand side of linear system). 
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               file containing the positions of EIT electrodes (.patches)
               output EITSourceOp
   -Head2EEGMat, -H2EM, -h2em : 
        Compute the linear application which maps the potential
        on the scalp to the EEG electrodes
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               file containing the positions of EEG electrodes (.patches)
               output matrix

   -Head2ECoGMat, -H2ECogM, -h2ecogm, -H2ECOGM : 
        Compute the linear application which maps the potential
        on the scalp to the ECoG electrodes
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               file containing the positions of ECoG electrodes (.patches)
               name of the interface on which to project the electrodes ("name")
               output matrix

   -Head2MEGMat, -H2MM, -h2mm : 
        Compute the linear application which maps the potential
        on the scalp to the MEG sensors
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               file containing the positions and orientations of the MEG sensors (.squids)
               output matrix

   -SurfSource2MEGMat, -SS2MM, -ss2mm : 
        Compute the linear application which maps the 
        distributed source  to the MEG sensors
            Arguments :
               mesh file for distributed sources (.tri .vtk .mesh .bnd)
               positions and orientations of the MEG sensors (.squids)
               output matrix

   -DipSource2MEGMat, -DS2MM, -ds2mm :  
        Compute the linear application which maps the current dipoles
        to the MEG sensors
            Arguments :
               dipoles positions and orientations
               positions and orientations of the MEG sensors (.squids)
               output matrix

   -Head2InternalPotMat, -H2IPM -h2ipm :  
        Compute the linear transformation which maps the surface potential
        and normal current to the value of the internal potential at a set of points within a volume
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               a mesh file or a file with point positions at which to evaluate the potential
               output matrix

   -DipSource2InternalPotMat, -DS2IPM -ds2ipm :   
        Compute the linear transformation  which maps the current dipoles
        to the value of the infinite potential at a set of points within a volume
            Arguments :
               geometry file (.geom)
               conductivity file (.cond)
               dipoles positions and orientations
               a mesh file or a file with point positions at which to evaluate the potential
               output matrix
               (Optional) domain name where lie all dipoles.

```

## Possible causes 
It is clear that we don't link proparly. We don't link the binaries statically, we don't ship all the needed libraries, and I think that this might have something to do with it

https://github.com/openmeeg/openmeeg/blob/c7d5d7481594c6f6d80c6803112bdf09c6b92d1f/OpenMEEG/cmake/FindMKL.cmake#L139

But either way I think we need to simplify lots of things from the building process, use more standard tools rather than custom `.cmake` files, and have much less hard coded paths. But probably I still don't have enough background to make informed decisions. 

### Final remarks
- look at the common question [here](https://cmake.org/Wiki/CMake_RPATH_handling) 
- we might be better off trying some sort of C++ package shipping ( like [conan](https://conan.io/) a pip for c++ packages) and build some recipies if needed. But let the openmeeg building as a clean barebone

massich:
The build does not link statically. It seems to be linked to `BUILD_SHARED_STATUS` variable. It seems that its value gets propagated to the other subprojects. 

https://github.com/openmeeg/openmeeg/blob/c7d5d7481594c6f6d80c6803112bdf09c6b92d1f/OpenMEEG/cmake/FindLibrary.cmake#L14-L15

But this does not happen. Even if you set up this variable to `OFF`, it gets reset to `ON`. I managed to build OpenMEEG subproject as static, by changing the status of this variable for the subproject, and build the subproject. However, I didn't found the way to propagate the right value from the parent project.

massich:
I think I found it. We where setting `USE_SYSTEM_library=0` instead of `BUILD_SHARED_LIB_library=OFF` in travis.

see this (and wait to see what travis has to say about it):
https://github.com/massich/openmeeg/commit/92bc2ad036962174535c0416c65016246ed84a0b

agramfort:
let me know how it goes...

massich:
[here's](https://travis-ci.org/massich/openmeeg/builds/312430183) a green travis

agramfort:
nice !

can you push directly to @papadop's PR?

Alex

massich:
@agramfort it is green 'cos I skipped VTK :) and mkl statically linked

agramfort:
that does the job for me :)

agramfort:
fixed by #258

PR #260, opened 2017-12-27
fix rpath using  on linux

agramfort:


agramfort:
#261 is more elegant. Closing

PR #261, opened 2017-12-27
fix rpath for tools on linux version 2

agramfort:
this is option 2 with global config...


agramfort:
now I need to understand how to do this with macos...

PR #262, opened 2017-12-28
fix rpath on mac to avoid setting DYLD_LIBRARY_PATH

agramfort:


agramfort:
works on mac too 🍻 !

agramfort:
for the record I copied the magic lines from:

https://github.com/kevoree/kevoree-cpp/blob/569eabd1b74c3dc002574da9e00d32ea1919531a/toolchain/osx.toolchain.cmake

PR #263, opened 2017-12-29
turn on mkl on appveyor

agramfort:
let's see how it goes....


massich:
I gess we can close this @agramfort

PR #269, opened 2018-03-19
[MRG] Add Openblas

massich:


massich:
@sagitter this should be in master already.

sagitter:
Yes, i have already successfully tested.

PR #275, opened 2018-03-21
[MRG] add OpenMEEG versions to the executables

massich:


agramfort:
that should work. Perfect !

massich:
@agramfort now is better, no need of creating own custom variables when CMake offers a standard way to do it.

agramfort:
LGTM !

massich:
Merging Now

PR #279, opened 2018-03-23
[WIP] Activate clang static analysis

massich:


agramfort:
@massich can you activate analyse ON on linux too so @papadop can see the errors it reports and eventually look into it?

massich:
I'm not sure about the usefulness of static analyzing during the generation of the building system. If we have it active, it complains about the `cpp` code that we use to test compiler functionalities. 
As example it complains that `res` is never used here: 
```cpp
int main() {
    double d = 3.0;
    const bool res = S::isnormal(d);
    return 0;
}
```

xx

massich:
The rest of the errors should be legid. @papadop are you taking over?

agramfort:
travis is green so no error?

massich:
There are plenty warnings. But no error. We can get it in, as it is and then improve the code when we start removing warnings.

I'll merge

PR #287, opened 2018-03-27
openblas on mac + llvm + go over readme

agramfort:


.travis.yml
```diff
@@ -119,14 +120,38 @@ before_install:
...
 script:
   - cd ${TRAVIS_BUILD_DIR}
   - mkdir -p build && cd build
   - |
     if [[ $BUILD_PACKAGE == "ON" && $TRAVIS_PULL_REQUEST == "false" && $TRAVIS_BRANCH == "master" ]]; then
-      export CMAKE_OPTIONS="-DCMAKE_CXX_STANDARD=11 -DBLA_VENDOR=$BACKEND"
+      export CMAKE_OPTIONS=" -DBLA_VENDOR=$BACKEND"
     else
-      export CMAKE_OPTIONS="-DCMAKE_CXX_STANDARD=11 -DBLA_VENDOR=$BACKEND -DENABLE_COVERAGE=ON"
+      export CMAKE_OPTIONS=" -DBLA_VENDOR=$BACKEND -DENABLE_COVERAGE=ON"
```
agramfort: @massich I remove this as set it in the cmake. no need to add this to cmkae line all the time.

massich: I disagree, I would rather take it out from cmake. But I can do it in a PR when adding specific compilation flags.

agramfort: pb is that cmake then does not work out of the box and you need cleverness to be able to build...

agramfort: ideally doing:

mkdir build
cd build
cmake .
make

should work provided you have installed the dependencies on your sytem with brew, conda or apt-get depending on your system. Cf. remark from @papadop about having good defaults for BLA_VENDOR

massich:
I understand. We should keep a list of things that are half way and take a
turn.
For this case, maybe we can do it outside of this PR with a commit message
with a keyword we can search for.

massich:
Secon thought adding it to cmake/options.cmake with `XXX` should be enough.
You can add it first thing in options or in root cmakelisrs so that we'll
find it often and we won't forget. Otherwise you can put it in
compiler_options which is where it should be, but is more likely to pass
unnoticed

PR #278, opened 2018-03-23
[MRG+1] Add python Bindings

massich:

PR #289, opened 2018-03-28
Add documentation

massich:


massich:
[Stackoverflow notes on SWIG and doxygen docstrings](https://stackoverflow.com/questions/37780325/generating-doxygen-comments-for-swig-generated-c-sharp-that-wraps-c)

massich:
the `doxy2siwg` command needs to be remade! 

the docstrings have never been used. The generated file `docstings.i` to import from **swig** is appended here `SWIG_MODULE_OPENMEEG_EXTRA_DEPS` and this varaible never used. (see this 5a4ef47a)


the entire wrapping needs a review. There are things like this variable `CMAKE_SWIG_FLAGS` that drive the generation process which was meant to be used by `UseSwig_AlwaysShared.cmake` but was removed in abcc0a25 and the variables never used.

massich:
`SWIG_MODULE_${MODULE_NAME}_EXTRA_DEPS` is an internal variable for `siwg_add_module`. So `SWIG_MODULE_XXXX_EXTRA_DEPS`should fit with `swig_add_module(XXXX python regular_file.i)`

massich:
I rebase master.

massich:
[fix BUILD_DOCUMENTATION with Python bindings cf #278](https://trello.com/c/iey753v9/101-fix-builddocumentation-with-python-bindings-cf-278)

agramfort:
merge when you're happy (enough)

thx @massich

documentation/reference/CMakeLists.txt
```diff
@@ -9,10 +9,22 @@ IF (BUILD_DOCUMENTATION)
...
+    #       The way to do that is:
+    #       find_package(Doxygen
+    #                    REQUIRED dot
+    #                    OPTIONAL_COMPONENTS gnuplot)
+    #       if(NOT DOXYGEN_DOT_FOUND)
+    #            message(WARNING "Graphviz doesn't seem to be installed.\
+    #                             Doxygen will not be able to generate graphs.\
+    #                             Consider installing this package.")
+    #       endif(NOT DOXYGEN_DOT_FOUND)
+    #
```
agramfort: shall we clean this?

documentation/CMakeLists.txt
```diff
@@ -1,7 +1,15 @@
-IF ( BUILD_DOCUMENTATION )
+# # Require dot, treat the other components as optional
+# find_package(Doxygen
+#              REQUIRED dot) # XXX if dot module if not present, cmake should stop generatio
+#              # OPTIONAL_COMPONENTS mscgen dia)
```
agramfort: here also?

documentation/reference/CMakeLists.txt
massich: No. This is here to point how this should be done. Either we change it and do it properly or add XXX comment, move on and let someone else pick it up (or Sunday afternoon).

massich: I mean

massich: + it should be done at least one layer above. (in your other [comment](https://github.com/openmeeg/openmeeg/pull/289/files/a5ccaffde385aecc7c909b170c9e2fba4933c54a#r178302874) documentation/CMakeLists.txt)

massich: But if we end up moving to fully documenting with Sphinx without doxygen, there would be no need for fixing it.

PR #298, opened 2018-04-09
[No MRG] Valgrind

massich:
Just a PR to check what valgrind does. 

Here follows the summary on my local machine:

```sh
Memory checking results:
Memory Leak - 302
Potential Memory Leak - 44568
        Errors while running CTest
```


massich:
closing this PR since it was [no-MRG] and we checked what we wanted. This can be reopen if needed. Maybe we should add a CI entry

PR #299, opened 2018-04-16
Add option to locally build libmatio

massich:


massich:
@agramfort ITS ALL GREEN !!!!!!!!

PR #301, opened 2018-04-17
[MRG] Enable to uninstall the package using: make uninstall

agramfort:


.travis.yml
```diff
@@ -45,7 +45,7 @@ matrix:
       - BACKEND=OpenBLAS
       # - USE_GCC=ON  # XXX : vtk does not work with GCC
       # - USE_BREW_LLVM=ON  # XXX: To have OpenMP but segfaults...
-      - USE_VTK=ON
+      # - USE_VTK=ON  # XXX : vtk packaging is currently broken with brew
```
agramfort: @massich I had to comment out the use of vtk on brew for now.

agramfort:
@massich merge if you're happy. thx

.travis.yml
```diff
@@ -197,6 +200,8 @@ script:
         PACKAGE_DEPLOY=$(greadlink -f OpenMEEG*.tar.gz)
       fi
     fi
+  - sudo make install
+  - sudo make uninstall
```
massich: Why those two ?

CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ include(cmake/CompilerOptions.cmake)
 include(cmake/options.cmake)
 include(cmake/thirdParties.cmake)
 include(cmake/install.cmake)  # install setup (do it before adding targets)
+include(cmake/Uninstall.cmake)  # install setup (do it before adding targets)
```
massich: do we need the uninstall? if so, we need I would like to double check it.

.travis.yml
agramfort: to test uninstall you need to install first.

CMakeLists.txt
agramfort: go for it. You should see that calling

make uninstall

removes all installed filed.

agramfort:
@massich can we move on here?

massich:
I don't understand the need for this. But if you think that having an uninstall target is useful, we should add the target itself. Which I think its not done here.

If I'm not mistaken, this PR only adds the helper functions to create the uninstall target. See [*uninstall target** in CMake FAQs](https://gitlab.kitware.com/cmake/community/wikis/FAQ#can-i-do-make-uninstall-with-cmake)

Which means, that 
```
- sudo make install
- sudo make uninstall
```
is not enough to test that we install and uninstall the package properly.

In my opinion, if a user compiles and installs a package, would manage to remove the package without the uninstall target.

massich:
So yes, we can merge!

.travis.yml
massich: I had to review, why we did that. (lets see how many times do I get to this comment)

PR #302, opened 2018-04-20
Remove VTK, since it's broken in BREW

massich:


agramfort:
@massich this will break. See my PR. vtk is installed with brew even if USE_VTK is not ON.

why not merging my PR?

massich:
in #301 there's all this thing about uninstall, etc.. I just did this one to have it in a cleaner manner.

You either merge this one or the other one, addressing the comments.

PR #306, opened 2018-06-06
[MRG] Report elapsed time based on wall-time

### LINKED ISSUE #294: Elapsed time does not correspond to the real time
See this log
```sh
~/code/random-thoughts/openmeeg_sample_data master* ⇡ 53s
❯ export OMP_NUM_THREADS=4                                

~/code/random-thoughts/openmeeg_sample_data master* ⇡
❯ time om_assemble -HM ${GEOMETRY} ${CONDUCTIVITIES} ${HM}
om_assemble version 2.4.0 compiled at Apr  4 2018 16:19:08 using OpenMP with 4 threads.


| ------ om_assemble
| -HM
| data/model/head_model.geom
| data/model/head_model.cond
| tmp24/tmp.hm
| -----------------------
This geometry is a NESTED geometry.
Info:: Mesh name/ID : Skull
		# vertices  : 842
		# triangles : 1680
		Euler characteristic : 2
		Min Area : 1.12847
		Max Area : 101.934
Info:: Mesh name/ID : Cortex
		# vertices  : 785
		# triangles : 1566
		Euler characteristic : 2
		Min Area : 5.45207
		Max Area : 88.4818
Info:: Mesh name/ID : Head
		# vertices  : 870
		# triangles : 1736
		Euler characteristic : 2
		Min Area : 3.09891
		Max Area : 206.874
Info:: Domain name : Scalp
		Conductivity : 1
		Composed by interfaces : +Skull -Head 
		Interface "Skull"= { mesh "Skull" }
		Interface "Head"= { mesh "Head"(outermost) }
Info:: Domain name : Brain
		Conductivity : 1
		Composed by interfaces : -Cortex 
		Interface "Cortex"= { mesh "Cortex" }
Info:: Domain name : Air
		Conductivity : 0
		Composed by interfaces : +Head 
		Considered as the outermost domain.
		Interface "Head"= { mesh "Head"(outermost) }
Info:: Domain name : Skull
		Conductivity : 0.03
		Composed by interfaces : +Cortex -Skull 
		Interface "Cortex"= { mesh "Cortex" }
		Interface "Skull"= { mesh "Skull" }
OPERATOR S ... (arg : mesh Skull , mesh Skull )
[********************]
OPERATOR D ... (arg : mesh Skull , mesh Skull )
[********************]OPERATOR N ... (arg : mesh Skull , mesh Skull )
[********************]
OPERATOR S ... (arg : mesh Cortex , mesh Skull )
[********************]
OPERATOR D ... (arg : mesh Cortex , mesh Skull )
[********************]OPERATOR D*... (arg : mesh Cortex , mesh Skull )
[********************]OPERATOR N ... (arg : mesh Cortex , mesh Skull )
[********************]
OPERATOR S ... (arg : mesh Cortex , mesh Cortex )
[********************]
[********************]OPERATOR N ... (arg : mesh Cortex , mesh Cortex )
[********************]
OPERATOR D*... (arg : mesh Head , mesh Skull )
[********************]OPERATOR N ... (arg : mesh Head , mesh Skull )
[********************]
[********************]
OPERATOR N ... (arg : mesh Head , mesh Head )
[********************]
[********************]
-------------------------------------------
| Elapsed Time: 92.6007 s.
-------------------------------------------
om_assemble -HM ${GEOMETRY} ${CONDUCTIVITIES} ${HM}  92.31s user 0.30s system 343% cpu 26.967 total
```

massich:
fixes #294


massich:
Now, elapsed time is Wall-time. See this trace:

```sh
~/code/random-thoughts/openmeeg_sample_data master* ⇡ 11s
❯ time om_assemble -HM ${GEOMETRY} ${CONDUCTIVITIES} ${HM}
om_assemble version 2.4.0 compiled at Jun  6 2018 11:49:39 using OpenMP
 Executing using 4 threads.


| ------ om_assemble
| -HM
| data/model/head_model.geom
| data/model/head_model.cond
| tmp/tmp.hm
| -----------------------
This geometry is a NESTED geometry.
Info:: Mesh name/ID : Skull
		# vertices  : 842
		# triangles : 1680
		Euler characteristic : 2
		Min Area : 1.12847
		Max Area : 101.934
Info:: Mesh name/ID : Cortex
		# vertices  : 785
		# triangles : 1566
		Euler characteristic : 2
		Min Area : 5.45207
		Max Area : 88.4818
Info:: Mesh name/ID : Head
		# vertices  : 870
		# triangles : 1736
		Euler characteristic : 2
		Min Area : 3.09891
		Max Area : 206.874
Info:: Domain name : Scalp
		Conductivity : 1
		Composed by interfaces : +Skull -Head 
		Interface "Skull"= { mesh "Skull" }
		Interface "Head"= { mesh "Head"(outermost) }
Info:: Domain name : Brain
		Conductivity : 1
		Composed by interfaces : -Cortex 
		Interface "Cortex"= { mesh "Cortex" }
Info:: Domain name : Air
		Conductivity : 0
		Composed by interfaces : +Head 
		Considered as the outermost domain.
		Interface "Head"= { mesh "Head"(outermost) }
Info:: Domain name : Skull
		Conductivity : 0.03
		Composed by interfaces : +Cortex -Skull 
		Interface "Cortex"= { mesh "Cortex" }
		Interface "Skull"= { mesh "Skull" }
OPERATOR S ... (arg : mesh Skull , mesh Skull )
[********************]
OPERATOR D ... (arg : mesh Skull , mesh Skull )
[********************]OPERATOR N ... (arg : mesh Skull , mesh Skull )
[********************]
OPERATOR S ... (arg : mesh Cortex , mesh Skull )
[********************]
OPERATOR D ... (arg : mesh Cortex , mesh Skull )
[********************]OPERATOR D*... (arg : mesh Cortex , mesh Skull )
[********************]OPERATOR N ... (arg : mesh Cortex , mesh Skull )
[********************]
OPERATOR S ... (arg : mesh Cortex , mesh Cortex )
[********************]
[********************]OPERATOR N ... (arg : mesh Cortex , mesh Cortex )
[********************]
OPERATOR D*... (arg : mesh Head , mesh Skull )
[********************]OPERATOR N ... (arg : mesh Head , mesh Skull )
[********************]
[********************]
OPERATOR N ... (arg : mesh Head , mesh Head )
[********************]
[********************]
-------------------------------------------
| Elapsed Time: 149.824 s.
-------------------------------------------
om_assemble -HM ${GEOMETRY} ${CONDUCTIVITIES} ${HM}  529.44s user 0.70s system 353% cpu 2:29.88 total
```

Now, the elapsed time reported by om_assemble and *nix time differ. 149 != 529.

agramfort:
LGTM. Can you see if we can remove old cpuChrono now?

massich:
@agramfort yes, that was missing.

`void dispEllapsed(const std::chrono::duration<double> elapsed_seconds)` is replicated code and it should be refactor. I think we should do that in a differnt PR and refactor the entire text-based interface at once.

apps/forward.cpp
```diff
@@ -102,3 +103,9 @@ void getHelp(char** argv)
 
     exit(0);
 }
+
+void dispEllapsed(const std::chrono::duration<double> elapsed_seconds){
+  std::cout <<  "-------------------------------------------" << std::endl;
+  std::cout <<  "| Elapsed Time: " << elapsed_seconds.count() << " s." << std::endl;
+  std::cout <<  "-------------------------------------------" << std::endl;
+}
```
agramfort: can you move this function in om_utils.h to avoid the code dupe?

massich: That's one option, its only half of the apps in this PR use `om_utils.h`. +1 to refactor the whole ncourses and unify all the apps. It's just that I'm not sure if it should be done here.

massich: And if we do it. Instead of `dispEllapsed`, we should provide iostring operator. So that we can write `std::cout << end_time-start_time;` (I think)

massich:
On the other side, I'm not sure if we should remove directly `cpuChrono.h` or to have a deprecation cycle. The right thing is the depreciation. But I'm not sure is worth it.

agramfort:
>
> On the other side, I'm not sure if we should remove directly cpuChrono.h
> or to have a deprecation cycle. The right thing is the depreciation. But
> I'm not sure is worth it.
>

don't bother about deprecating. OpenMEEG is so far not used as a library.

massich:
[Modify elapse time to be real time](https://trello.com/c/G721ML4b/115-modify-elapse-time-to-be-real-time)

agramfort:
LGTM !

+1 for MRG when green.

papadop:
OK for me (I'll push the merge button). But please in general avoid the #include "" form. Prefer the <> form. Anyhow, these lines will be removed sooner or later, so I do not bother much here. And for the record I agree about having the iostream version....

agramfort:
the logic we (will) follow is:

use "" for local includes
use <> for include from a used library.

it seems a standard in the C++ world these days.

agramfort:
also @papadop please wait for CIs to go green before merging. It's still running on Mac:

https://travis-ci.org/openmeeg/openmeeg_packages/jobs/388761646

papadop:
Thank you for the C course ;-) (It is actually an old K&R rule, which is kept in C++ for backward compatibility, I have the original book -- even before C89 -- K&R if you want).
The thing is that most of the time the distinction is subtle and strongly depends on whether the file will be installed or not.
And in my experience, most of the time it ends being wrong at one or another time (because the file is moved, because we change our minds and it is installed, ...).
Using <> work as well and has the small cost of adding the current directory in cmake (which we do), is visually more consistent, and less error prone in the end....

papadop:
Strange in my interface it was saying "all green"... The two builds (which I interpreted as travis and appveyor) where shown as successful....

agramfort:
:)

I would say let's stick to a rule we can agree on. We all know that rules
are generally imperfect

I see:

"" for local includes ?

"" for headers that not installed?

papadop:
Another note is that there seems to be some work ahead to migrate some github service:


Compare: https://github.com/openmeeg/openmeeg/compare/3a77237ca532...a6373838aef7
      **NOTE:** This service been marked for deprecation: https://developer.github.com/changes/2018-04-25-github-services-deprecation/

      Functionality will be removed from GitHub.com on January 31st, 2019.

agramfort:
disseminating good practices and standards?

massich:
Could we revert and Squash-Merge the PR? It keeps the history much cleaner. 


Regarding <> / "" / installing, etc.. There's no such thing as: 
> (because the file is moved, because we change our minds and it is installed, ...)

They should always be associated with tests, so that they can not be changed without modifying the tests. 

So at some point, we should agree on those things make it strict and allow a bot to keep track for ourselfs.

I settit up with "" 'cos this file should defenitely go away.

papadop:
It is because the file will go away that I pushed the merge button, otherwise I would have been more intolerant  ;-)
Why asking people to change things in three places (CMake, test and code) when we can avoid it and focus on making a change in one single place ??
That's not good practise, that's putting annoying rules... just to use a poorly designed old C rule (I can show you threads and threads about people misunderstanding the feature). And that with no real advantage (in the old C days I think it was mainly made to somewhat speed-up compilation and maybe give some control to mask a system include by a local one, honestly I do not see what this brings us).

The only use I'd agree to would be the one you used "We want to remove this file"... But a comment would be much clearer...

agramfort:
I am not dogmatic but if we could have a page that list the guidelines we
follow
in this project it would be great.

You know that I asked already many times that we use a linter.

PR #309, opened 2018-06-07
[MRG] Fix missing zlib in windows

massich:
Do not look at this.


massich:
All the trouble seems to be coming from here:
https://github.com/openmeeg/openmeeg/blob/master/OpenMEEGMaths/include/MatlabIO.H#L322

agramfort:
see
https://github.com/openmeeg/openmeeg/blob/master/cmake/Findmatio.cmake#L25
that I had to do on mac to make sure libz static is used.

massich:
@openmeeg/owners this is good on my end. **Squash and merge** when ready.

massich:
Does the ping to owners work?
cc: @agramfort @papadop

agramfort:
Great job @massich

PR #333, opened 2018-08-03
try using dist: precise

agramfort:
objective is use older glibc versions...


agramfort:
we have 2 pbs: cmake is too old on this system and MKL version we use is too recent :)

PR #337, opened 2018-08-16
[BUG] [CI] mkl installation breaks in MACOSX 10.13

massich:


agramfort:
I get the pb. MKL needs to be updated if you update mac os x version. We'll fix this when we need to.

PR #338, opened 2018-08-17
WIP:Use patched matio from massich conda channel

massich:


massich:
close in favor of #339

Issue #341, opened 2018-08-17
Degenerated gain matrix when using om_gain


massich:
here's the execution with [this data](https://www.dropbox.com/sh/av40f6brhv5qdxn/AAApYZbe8eC-3-YCV74AvUXNa?dl=0)

```sh
./om_gain -EEG \
          ~/Dropbox/openmeeg_segfault/openmeeg_hminv.mat \
          ~/Dropbox/openmeeg_segfault/openmeeg_dsm.mat \
          ~/Dropbox/openmeeg_segfault/openmeeg_h2em.mat \
          /tmp/gain_eeg.mat

om_gain version 2.4.0 compiled at Aug 17 2018 15:21:43 using OpenMP
 Executing using 4 threads.


| ------ om_gain
| -EEG
| ./openmeeg_hminv.mat
| ./openmeeg_dsm.mat
| ./openmeeg_h2em.mat
| /tmp/openmeeg_gain_eeg.mat
| -----------------------
Using variable with name: matrix
-------------------------------------------
| Elapsed Time: 21.4012 s.
-------------------------------------------

~/Dropbox/openmeeg_segfault 21s
❯ om_matrix_info -i /tmp/gain_eeg.mat                                                       
om_matrix_info version 2.4.0 compiled at Aug 17 2018 15:22:01 using OpenMP
 Executing using 4 threads.

Loading : /tmp/gain_eeg.mat
Dimensions : 257 x 45000
Min Value : -330.452 (124,1212)
Max Value : 170.216 (124,1192)
First Values
13.0735 -9.86959 0.198685 13.0337 -9.88526 
16.1424 -9.24023 10.517 16.2927 -9.54824 
15.7827 -7.38458 15.9536 15.9731 -7.73443 
14.5447 -5.27728 18.6208 14.7053 -5.55507 
12.9091 -3.77995 20.3059 13.0044 -3.98566 
```
The operation seems to go fine but loading the data with matlab shows a degenerated gain matrix:

```matlab
>> load /tmp/openmeeg_gain_eeg.mat
>> figure; imagesc(linop);colorbar
```
![image](https://user-images.githubusercontent.com/7044835/44289573-2d51a980-a275-11e8-944e-8de4efa92886.png)

massich:
the results are replicated here with a different data set https://neuroimage.usc.edu/forums/t/error-while-computing-sources-input-to-svd-must-not-contain-nan-of-inf/7806/11

agramfort:
[...]

PR #342, opened 2018-08-20
[WIP] attempt to use mat73 for IO

agramfort:
Let's see what breaks with Matlab 73 format...

### DIFF
--- a/OpenMEEGMaths/include/MatlabIO.H
+++ b/OpenMEEGMaths/include/MatlabIO.H
@@ -215,7 +215,7 @@ namespace OpenMEEG {
                     os.close();
                     std::remove(name().c_str());
                 }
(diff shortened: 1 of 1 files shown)


papadop:
All checks are passing.... So what were the problems with matlab 7.3 file format and windows 64 bits ???

agramfort:
Yes good surprise ... maybe we were at that time not using the conda package. 

@massich can you check if this works with the files on Dropbox ?

agramfort:
ok I am going to merge this as CIs are green. I want the packaged binaries to be able to test.

agramfort:
erratum. It does work.

massich:
this fixes #341 indeed.
image

we are bumping the version to 2.4.1 #344 with the fixes.

PR #344, opened 2018-08-22
[MRG] bump version to release last fixes

massich:


massich:
@agramfort merge when green and let's see how much things we break when doing so!

PR #345, opened 2018-08-23
[MRG] Bump master to 2.5.dev0

massich:


massich:
lets see how this works. If it does, then we should update whats new, etc.. in master and 2.4.x branches.

also tags, tars, packages, etc...

massich:
https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html

hichamjanati:
https://cmake.org/cmake/help/latest/command/project.html

PR #354, opened 2019-03-14
FIX: make sure meg sensors as numbers with integration points are still ok

### LINKED ISSUE #353: Change of behavior of om_gain: it used to apply weights
The function om_gain used to return only one row per sensor, no matter how many integration points are used for these points. Now it returns one row per integration point.

In the example below, I have processed the same CTF example with 9 magnetometers (4 integration points) and 291 gradiometers (8 integration points) from Brainstorm with two versions of OpenMEEG:
https://www.dropbox.com/s/q1e9dfvalmz5hwo/openmeeg-2.1.zip?dl=0
https://www.dropbox.com/s/7m4y85ojxluk2a0/openmeeg-2.4.zip?dl=0

OpenMEEG 2.1: openmeeg_gain_meg.mat = [300 x 3009]
OpenMEEG 2.4: openmeeg_gain_meg.mat = [2364 x 3009]

This causes the current Brainstorm wrapper not to work anymore, as there is nothing to apply the weights after the call to om_gain:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/forward/bst_openmeeg.m#L458

And therefore users complaining:
https://neuroimage.usc.edu/forums/t/issue-forward-modelling-with-openmeeg-bem/10266/
https://neuroimage.usc.edu/forums/t/error-in-openmeeg-in-meg-data/8421

Is this related to this other issue?
https://github.com/openmeeg/openmeeg/issues/347

Did this become an optional parameter controlled with a command line flag?
Is it something that you will fix, or something that should be handled in the Brainstorm side instead? If so, what should we do: simply sum the rows multiplied by the weights?

agramfort:
closes #353 

cc @ftadel

@papadop @mclerc this commit https://github.com/openmeeg/openmeeg/commit/cbb9d0f574c32d40fedd12fd8489d3cdd74fdbb4 had broken brainstorm in 2014.

@massich can you see if you can cut a new minor release? also we need a what's new page. @massich can you help with this? thanks a lot

### DIFF
--- a/OpenMEEG/src/sensors.cpp
+++ b/OpenMEEG/src/sensors.cpp
@@ -90,7 +90,6 @@ namespace OpenMEEG {
         std::string s, buf;
         Strings names;
         Strings tokens;
-        Strings::const_iterator tokensIterator;
         bool labeled = false;
         size_t nlin = 0;
         size_t ncol = 0;
--- a/OpenMEEG/src/sensors.cpp
+++ b/OpenMEEG/src/sensors.cpp
@@ -107,7 +107,9 @@ namespace OpenMEEG {
                         ncol++;
                     }
                 }
-                tokensIterator = tokens.begin();
+                if (ncol >= 7) {
+                    labeled = true;
+                }
                 for ( size_t j = 0; j < tokens[0].size(); ++j) {
                     if ( isalpha(tokens[0][j]) && (tokens[0][j] != 'e') && (tokens[0][j] != 'E' ) ) {
                         labeled = true;


papadop:
Looking at the code and the squids files in the source code, it seems that the format has somewhat changed (while trying to keep backward compatibility). It also looks that there are no tests (at least MEG ones) with integration and that's why  we have not caught the problem.

In the old format, sensor names were allowed to be numbers. In the new one they should start by a letter (and not a 'e' or 'E' if I'm looking in the code which is strange and contradicts the documentation: I need to investigate more). I need to understand better why this constraints has been added but I guess this is related to the addition of EIT sensors and to avoid some ambiguity. A quick guess is that in the old format, we were "guessing" the type of sensor based on the number of columns which is not longer possible with EIT sensors. The constraint is a way to maintain the old format and its heuristic way of determining the sensor type, but I need to investigate more.

For now a simple fix is to add a letter (not 'e' or 'E') in front of any numerical label. I'm pretty sure that Francois files were not interpreted as MEG channels... I'm not sure there is a good fix that would allow for the new functionality while preserving the use of the old format. I'll try to come with a solution, but surely, there is the need to change the file format and stop to rely on an heuristics to determine the sensor type.

This is a quick analysis, I might have missed something. More is coming later.

agramfort:
yes it's true that if you prefix the channel number with 'M' it works with
current release version.
Yes we need a test and make sure the documentation is correct.

@papadop do you have time to help here?

papadop:
As I said, I'll try. In any case, it will be within the next week or then in a month. But I'll try to do some (fun) "technical work" to distract myself from the annoying stuff I have to do.... I'll probably will have a look later tonight or tomorrow.

papadop:
Yet the main decision we have to take is what to do... It is clear to me, that maintaining the full compatibility with the old mode will be difficult... Heuristics and implicit information cannot work on the long run. My guess is that unfortunately there will be some breakage...

So the first question is what format should we adopt. And I propose that whatever is that format, we have a clear way to identify it and - at the very least - emit a warning or a clear error message. In the best scenario (but I do not promise that yet), we could keep backward compatibility for a while....

As for the format, it does not need to be very different from the one currently implemented, but we could require a text header allowing to detect the new format... or we could require to have a keyword per sensor to give its type. I need to think a little about it, but ideas are welcome....

agramfort:
BIDS is a new standard and is based on TSV files so maybe we don't need to
invent anything. What I don't understand is why it needed to be broken to
support EIT?

>

ftadel:
I confirm that replacing the sensor name from a simple index "1" to a text label "MEG001" fixes the issue:
https://github.com/brainstorm-tools/brainstorm3/commit/7e463aa1eff95a86c0f3e6208a364ed2ebdf99f3#diff-9e590703ce3a61a47c365d821b1164e6

papadop:
Because the "type" of channel is implicit and deduced by the number of columns (or I should say numerical columns). This is deducible from the documentation at the beginning of sensors.h. Contrarily to what I said, this is not really the type (EEG, MEG or EIT) but the variant of the type and more precisely the "patch" version which allows to define "circular" measurements area around a point. For EEG and EIT these are defined with a position and a radius which conflict with a number describing the channel and a position. This is where the constraint of having names starting with a character comes from... There is probably a similar ambiguity for MEG where it was possible to have either 6 coefficients preceded  by a name or just 6 coefficients and a weight... Again the character rule allows to remove the ambiguity.

I do not know how the problem was avoided before... Either, the code was wrong. Or there were no weights (which seems strange since I guess that how we implement gradiometers), or ??? I need to compare with the old version.

agramfort:
we have a bug fix in brainstorm already. no rush here then.

papadop:
As for BIDS or TSV, I have no clue... Are there available and simple to use parsers  ? Does the format contains all what we need ? I am a little afraid that BIDS is quite complicated (json and other stuff like this), which I'm not very keen to introduce without studying very carefully the implications. In any case, while it probably can come as a new alternative type of input, it cannot replace the existing one immediately. On the other hand, adding a BIDS2OpenMEEG script (or even an OpenMEEG2BIDS), might be nice and not that difficult since there seems to be a python reference code.

agramfort:
channel info is not in json files but plain .tsv

but if we finally manage to have good python bindings then MNE will do all
the hard work and will just pass numpy arrays in memory to openmeeg. it's
for me the best solution so openmeeg stays focused on solving boundary
integral equations.

massich:
I agree that the easiest and quickest solution here is https://github.com/brainstorm-tools/brainstorm3/commit/7e463aa1eff95a86c0f3e6208a364ed2ebdf99f3#diff-9e590703ce3a61a47c365d821b1164e6

But this does not drop the fact that we need a thorough test suit in openmeeg validating inputs. We keep finding things like this one, and this should not really be the case.

agramfort:
+1 to add a test in this PR and revert my change and just document the API
change with new 2.4 version

>

papadop:
For the record, since I do not think I will be able to make a change before at least two weeks. I analyzed the code more carefully. The intent of the code is to allow for everything but a floating point for the label. Unfortunately, the code does not really attempt to parse a full floating point hence this strange rule about 'E' and 'e' (exponents of floating point numbers). The most conservative fix would be to try to parse a real floating point (i.e. containing a . and/or an exponent) and to reject only in that case. A pure integer would not be recognized as a floating point (which is awkward but necessary to restore the old way it worked).

I really think that  a better fix would be to revisit this file format. But since we do not want to break backward compatibility we may have to implement the above anyhow. Another question is why do we ask for homogeneous sensor files (one type of specification per file, including the format specification ( for example MEG sensors specified with 3 (position) or 6 (position+orientation) parameters must be in different files !!!! While I agree that this covers most real cases, it is a nonsense...

agramfort:
sleeping over it I would just close this PR. there is a simple workaround
and I don't want to inject more complexity in openmeeg.

>

papadop:
I do not think it is wise. We should at the very least strengthen the code and add a testcase. This is not very urgent though, since indeed the fix is quite simple.

agramfort:
ok for me

>

eolivi:
I agree with @papadop .
The tricky point was to decide between a float, an int, or a string (for the first column).
This check was a (bad) hack for this.
It would be so much easier using boost::lexical_cast https://stackoverflow.com/questions/8065413/stdlexical-cast-is-there-such-a-thing
But as papadop said, we should check whether it is a float (and not an int), if yes: {it is the x coordinate } if not => it is a label

massich:
If I recall correctly we didn't have a boost dependency so far.

eolivi:
no we don't have boost

massich:
I think we can close this one for the moment, don't we?

PR #356, opened 2019-03-24
fix CMAKE_BUILD_TYPE default value

eolivi:
if you don't set CMAKE_BUILD_TYPE it remains void and does not compile with "-O3" optimization, ...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,25 +3,22 @@ project (OpenMEEG VERSION 2.4.9999 LANGUAGES C CXX)
 
 set(CMAKE_CXX_STANDARD 11)  # use c++11
 
-set(OpenMEEG_HEADER_INSTALLDIR include) # XXX I don't know why this should be here
-
 # Modify where to find the modules
 list(APPEND CMAKE_MODULE_PATH
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake/utils
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake/FortranCInterface
-    ${CMAKE_MODULE_PATH}
 )
 
 ################
 # OpenMEEG Configuration and needed things
 ################
 include(GenerateExportHeader)
-include(cmake/CompilerOptions.cmake)
-include(cmake/options.cmake)
-include(cmake/thirdParties.cmake)
-include(cmake/install.cmake)  # install setup (do it before adding targets)
-include(cmake/Uninstall.cmake)  # install setup (do it before adding targets)
+include(CompilerOptions)
+include(options)
+include(thirdParties)
+include(install)  # install setup (do it before adding targets)
+include(Uninstall)  # install setup (do it before adding targets)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/OpenMEEGConfigure.h.in
                ${CMAKE_CURRENT_BINARY_DIR}/OpenMEEGConfigure.h)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -3,25 +3,22 @@ project (OpenMEEG VERSION 2.4.9999 LANGUAGES C CXX)
 
 set(CMAKE_CXX_STANDARD 11)  # use c++11
 
-set(OpenMEEG_HEADER_INSTALLDIR include) # XXX I don't know why this should be here
```
eolivi: I think we can remove that... but not sure

cmake/thirdParties.cmake
```diff
@@ -4,6 +4,9 @@
 
 set(BLA_DEFINITIONS)
 
+set(BLA_VENDOR "OpenBLAS" CACHE STRING "BLAS/LAPACK implementation")
```
agramfort: this overwrites argument passed by user?

cmake/thirdParties.cmake
```diff
@@ -4,6 +4,9 @@
...
+# cannot set this as travis sometimes use Intel10_64lp ...
+#set_property(CACHE BLA_VENDOR PROPERTY STRINGS OpenBLAS Intel ATLAS Apple)
```
agramfort: to remove then?

agramfort:
@massich can you also have a look? including to why travis fails on any new PR?

thanks a lot @eolivi

cmake/thirdParties.cmake
eolivi: no it does not (if you want to override you need FORCE)
it set a default value if there were no value

cmake/thirdParties.cmake
eolivi: yes... unless you want to turn your travis variable Intel10_64lp into Intel (other cases ?...)

cmake/thirdParties.cmake
massich: if I'm not mistaken, if not set it was complaining to force the user to make a choice.

CMakeLists.txt
```diff
@@ -3,25 +3,22 @@ project (OpenMEEG VERSION 2.4.9999 LANGUAGES C CXX)
...
 ################
 # OpenMEEG Configuration and needed things
 ################
 include(GenerateExportHeader)
-include(cmake/CompilerOptions.cmake)
-include(cmake/options.cmake)
-include(cmake/thirdParties.cmake)
-include(cmake/install.cmake)  # install setup (do it before adding targets)
-include(cmake/Uninstall.cmake)  # install setup (do it before adding targets)
+include(CompilerOptions)
```
massich: I am -1 in using only the file name. Explicit is better than implicit, and if your cmake paths are messed up (like having duplicated files in `cmake_old` for historical purposes) then it's not always clear.

cmake/thirdParties.cmake
massich: thinking about it I'm also -1 for this one. Setting `BLA_VENDOR` to something that you don't have causes a more nonesense error down the line.

it forces the user to set it.

Maybe what we should do is to keep it this way (unset) and write a nicer error message asking the user to take the time to think what does he want to compile and pass it properly. 

IMHO not setting things so that they carsh on the right time or force to pass parameters, is a feature not a bug.

agramfort:
is there maybe just a way to display the possible options as error message?

>

eolivi:
take whatever you want, I let you decide @massich ;-)

massich:
Thx @eolivi

PR #360, opened 2019-03-29
Fix: gforge upload

massich:


massich:
great it fails in the same manner

PR #362, opened 2019-04-06
New sensor classes specialized for EEG, MEG, EIT, ECoG.

eolivi:
I did this, it is not over yet.
I think we should go toward something like this, to avoid hacks like we did to guess what type of sensors are being parsed.

Let me know what you think.

WIP: since it is probably not tested enough, and some function need to be written still (save(), ...)

@papadop @agramfort @massich

### DIFF
--- a/OpenMEEG/include/MEGsensors.h
+++ b/OpenMEEG/include/MEGsensors.h
@@ -76,12 +76,12 @@ namespace OpenMEEG {
 
     class OPENMEEG_EXPORT MEGSensors : public Sensors {
     public:
-        MEGSensors(): Sensors() { }; /*!< Default constructor. */
-        MEGSensors(const char* filename): Sensors() { load(filename); }; /*!< Construct from file. */
+        MEGSensors(): Sensors("MEG") { }; /*!< Default constructor. */
+        MEGSensors(const char* filename): Sensors("MEG") { load(filename); }; /*!< Construct from file. */
 
         void info(int n_lines = 5) const; /*!< \brief get n_lines first lines info about sensors. */
         void load(const char* filename); /*!< Load sensors from file. */
-        void save(const char* filename);
+        void save(const char* filename) const;
         Matrix& getOrientations() {return m_orientations ; } /*!< Return a reference on sensors orientations. */
         Matrix getOrientations() const {return m_orientations ; } /*!< Return a copy of sensors orientations. */
         Vector getOrientation(size_t idx) const; /*!< Return the orientations (3D point) of the integration point i. */
--- a/OpenMEEG/include/MEGsensors.h
+++ b/OpenMEEG/include/MEGsensors.h
@@ -1,0 +1,102 @@
+/*
+Project Name : OpenMEEG
+
+© INRIA and ENPC (contributors: Geoffray ADDE, Maureen CLERC, Alexandre
+GRAMFORT, Renaud KERIVEN, Jan KYBIC, Perrine LANDREAU, Théodore PAPADOPOULO,
+Emmanuel OLIVI
+Maureen.Clerc.AT.inria.fr, keriven.AT.certis.enpc.fr,
+kybic.AT.fel.cvut.cz, papadop.AT.inria.fr)
+
+The OpenMEEG software is a C++ package for solving the forward/inverse
+problems of electroencephalography and magnetoencephalography.
+
+This software is governed by the CeCILL-B license under French law and
+abiding by the rules of distribution of free software.  You can  use,
+modify and/ or redistribute the software under the terms of the CeCILL-B
+license as circulated by CEA, CNRS and INRIA at the following URL
+"http://www.cecill.info".
+
+As a counterpart to the access to the source code and  rights to copy,
+modify and redistribute granted by the license, users are provided only
+with a limited warranty  and the software's authors,  the holders of the
+economic rights,  and the successive licensors  have only  limited
+liability.
+
+In this respect, the user's attention is drawn to the risks associated
+with loading,  using,  modifying and/or developing or reproducing the
+software by the user in light of its specific status of free software,
+that may mean  that it is complicated to manipulate,  and  that  also
+therefore means  that it is reserved for developers  and  experienced
+professionals having in-depth computer knowledge. Users are therefore
+encouraged to load and test the software's suitability as regards their
+requirements in conditions enabling the security of their systems and/or
+data to be ensured and,  more generally, to use and operate it in the
+same conditions as regards security.
+
+The fact that you are presently reading this means that you have had
+knowledge of the CeCILL-B license and that you accept its terms.
+*/
+
+#pragma once
+
+#include <vector>
+
+#include <IOUtils.H>
+#include <vector.h>
+#include <matrix.h>
+#include <geometry.h>
+#include <sensors.h>
+
+#include <OpenMEEG_Export.h>
+
+namespace OpenMEEG {
+
+    /*!
+     *  Sensors class for EEG and MEG sensors.
+     *  This class is made for reading sensors description file. This description file is a file text. Sensors may have names (labels)
+     *  in the first column of the file (it has to contains at least one character to be considered as label)
+     *  the file can have the shape of (neglecting if present the first, label column):
+     *  <ul>
+     *
+     *  <li> 1 line per sensor and 3 columns (MEG sensors without orientation)
+     *        <ul TYPE="circle">
+     *        <li> the 1st, 2nd and 3rd columns are respectively position coordinates x, y, z of sensor  </li>
+     *        </ul>
+     *  </li>
+     *  <li> 1 line per sensor and 6 columns (MEG sensors) :
+     *        <ul TYPE="circle">
+     *        <li> the 1st, 2nd and 3rd are respectively position coordinates x, y, z of sensor  </li>
+     *        <li> the 4th, 5th and 6th are coordinates of vector orientation </li>
+     *        </ul>
+     *  </li>
+     *  <li> 1 line per integration point for each sensor and 7 columns (MEG sensors) :
+     *        <ul TYPE="circle">
+     *        <li> the 1st, 2nd and 3rd are respectively position coordinates x, y, z of sensor  </li>
+     *        <li> the 4th, 5th and 6th are coordinates of vector orientation </li>
+     *        <li> the 7th is the weight to apply for numerical integration (uses sensor name) </li>
+     *        </ul>
+     *  </li>
+     *  </ul>
+     */
+
+    class OPENMEEG_EXPORT MEGSensors : public Sensors {
+    public:
+        MEGSensors(): Sensors() { }; /*!< Default constructor. */
+        MEGSensors(const char* filename): Sensors() { load(filename); }; /*!< Construct from file. */
+
+        void info(int n_lines = 5) const; /*!< \brief get n_lines first lines info about sensors. */
+        void load(const char* filename); /*!< Load sensors from file. */
+        void save(const char* filename);
+        Matrix& getOrientations() {return m_orientations ; } /*!< Return a reference on sensors orientations. */
+        Matrix getOrientations() const {return m_orientations ; } /*!< Return a copy of sensors orientations. */
+        bool hasOrientations() const { return m_orientations.nlin() > 0 ;} /*!< Return true if contains orientations */
+        Vector getOrientation(size_t idx) const; /*!< Return the orientations (3D point) of the integration point i. */
+        void setOrientation(size_t idx, Vector& orient); /*!< Set the orientation (3D point) of the integration point i. */
+
+        Vector getWeights() const { return m_weights; }
+        SparseMatrix getWeightsMatrix() const;
+    private:
+        Vector m_weights;                   /*!< Weights of integration points */
+        Matrix m_orientations;              /*!< Matrix of sensors orientations. ex: orientation(i,j) with  j in {0,1,2} for sensor i */
+    };
+}
--- a/OpenMEEG/include/MEGsensors.h
+++ b/OpenMEEG/include/MEGsensors.h
@@ -58,11 +58,6 @@ namespace OpenMEEG {
      *  the file can have the shape of (neglecting if present the first, label column):
      *  <ul>
      *
-     *  <li> 1 line per sensor and 3 columns (MEG sensors without orientation)
-     *        <ul TYPE="circle">
-     *        <li> the 1st, 2nd and 3rd columns are respectively position coordinates x, y, z of sensor  </li>
-     *        </ul>
-     *  </li>
      *  <li> 1 line per sensor and 6 columns (MEG sensors) :
      *        <ul TYPE="circle">
      *        <li> the 1st, 2nd and 3rd are respectively position coordinates x, y, z of sensor  </li>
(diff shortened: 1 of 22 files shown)


agramfort:
new strong objection here. Tell us when we can review @eolivi

massich:
IMHO before attacking this kind of issues, we could do things like:
- unifying argparse
- migrate to a moder testsuit
- change the matrix backend to a thirdparty that we can pass the matrices to python without writting them to disc

but I might be completely off

agramfort:
I agree that if we start thinking of actually moving forward we need to
discuss priorities.
I have no strong feeling about this PR as it does not break user's code.
I don't think that changing our file format for sensors is a top priority.

my 2c

>

eolivi:
This PR is almost ready. We have a (minor) choice to do:
- 1) either: I do Sensors class abstract, and we inherit for all EEG, MEG, EIT, ECoG Sensor from it
- 2) or: Sensors class is the EEG class and others derive from it

1) in the first case, we would need to change all the python calls like sensors= om.Sensors() to sensors=om.EEGSensors() , etc..
2) we only need in our/your python scripts to change the MEG, EIT, ECoG sensors, since the EEG case would work as before (since EEGSensors() = Sensors())

tell me your preference: @agramfort voted for the (1).

massich:
I think 1 is better.

agramfort:
option 1 will be easier to document. EEG is no special case.

>

eolivi:
it can be reviewed I guess, but please take your time.
For ex: I don't remember any specificities for the ECoG sensors are there any difference from an EEG sensor ?

OpenMEEG/include/MEGsensors.h
```diff
@@ -0,0 +1,102 @@
...
+    public:
+        MEGSensors(): Sensors() { }; /*!< Default constructor. */
+        MEGSensors(const char* filename): Sensors() { load(filename); }; /*!< Construct from file. */
+
+        void info(int n_lines = 5) const; /*!< \brief get n_lines first lines info about sensors. */
+        void load(const char* filename); /*!< Load sensors from file. */
+        void save(const char* filename);
+        Matrix& getOrientations() {return m_orientations ; } /*!< Return a reference on sensors orientations. */
+        Matrix getOrientations() const {return m_orientations ; } /*!< Return a copy of sensors orientations. */
+        bool hasOrientations() const { return m_orientations.nlin() > 0 ;} /*!< Return true if contains orientations */
```
agramfort: It cannot be False for an MEG sensor. I guess the hasOrientations method is now unecessary.

OpenMEEG/src/MEGsensors.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    void MEGSensors::setOrientation(size_t idx, Vector& orient) {
+        return m_orientations.setlin(idx,orient);
+    }
+
+    void MEGSensors::load(const char* filename) {
+        // line with
+        // 3 elements = position
+        // or
+        // 4 elements = label + position
```
agramfort: this cannot be. You must have orientations for MEG

agramfort:
@eolivi looking at coverage https://codecov.io/gh/openmeeg/openmeeg/compare/1fc56ec4819af81f94ad5339cce5e15534a4c3a7...465a1202e00fac34354bdc3abbc8ff6cd658cf47/tree?path=OpenMEEG%2Fsrc (which is not great) can you see if a test can be udpated to improve it? thanks a lot

eolivi:
I could use tests/test_sensors.cpp but this folder is not anymore used ?? @massich  
The only running tests are in apps/ now, right ? does it cover non regression tests still ?

wrapping/src/test_python.py
```diff
@@ -51,11 +53,21 @@
...
 dipoles.load(dipole_file)
 
-sensors = om.Sensors()
+sensors = om.MEGSensors()
 sensors.load(squidsFile)
+sensors.info()
 
-patches = om.Sensors()
+patches = om.EEGSensors()
 patches.load(patches_file)
```
agramfort: since we create new classes here. Would it make sense to be able to pass the filename to the constructor so we would just write:

`patches = om.EEGSensors(patches_file)`

?

OpenMEEG/include/MEGsensors.h
```diff
@@ -0,0 +1,96 @@
...
+
+    class OPENMEEG_EXPORT MEGSensors : public Sensors {
+    public:
+        MEGSensors(): Sensors() { }; /*!< Default constructor. */
+        MEGSensors(const char* filename): Sensors() { load(filename); }; /*!< Construct from file. */
+
+        void info(int n_lines = 5) const; /*!< \brief get n_lines first lines info about sensors. */
+        void load(const char* filename); /*!< Load sensors from file. */
+        void save(const char* filename);
+        Matrix& getOrientations() {return m_orientations ; } /*!< Return a reference on sensors orientations. */
```
massich: this is not cover

OpenMEEG/include/MEGsensors.h
```diff
@@ -0,0 +1,96 @@
...
+        Matrix getOrientations() const {return m_orientations ; } /*!< Return a copy of sensors orientations. */
+        Vector getOrientation(size_t idx) const; /*!< Return the orientations (3D point) of the integration point i. */
+        void setOrientation(size_t idx, Vector& orient); /*!< Set the orientation (3D point) of the integration point i. */
+
+        Vector getWeights() const { return m_weights; }
+        SparseMatrix getWeightsMatrix() const;
+    private:
+        Vector m_weights;                   /*!< Weights of integration points */
+        Matrix m_orientations;              /*!< Matrix of sensors orientations. ex: orientation(i,j) with  j in {0,1,2} for sensor i */
+    };
```
massich: this is not cover

OpenMEEG/include/sensors.h
```diff
@@ -92,74 +93,85 @@ namespace OpenMEEG {
...
+                while (iss >> buf) {
+                    tokens.push_back(buf);
+                }
+                // it is labeled unless there exists a float (i.e containing one '.')
+                if (std::count(tokens[0].cbegin(), tokens[0].cend(), '.') == 1) {
+                    seems_labeled = false;
+                }
+                if (nlin == 0) // first line
+                    ncol = tokens.size();
+                else if ( tokens.size() != ncol ) {
```
massich: this is not cover

OpenMEEG/include/sensors.h
```diff
@@ -92,74 +93,85 @@ namespace OpenMEEG {
...
-    }
-
     inline void Sensors::setPosition(size_t idx, Vector& pos) {
         return m_positions.setlin(idx,pos);
     }
+}
 
-    inline void Sensors::setOrientation(size_t idx, Vector& orient) {
-        return m_orientations.setlin(idx,orient);
+namespace {
```
massich: I don't think that code should be placed in the header. There's no need to remove `sensors.cpp` and let there whatever belongs there.

OpenMEEG/include/sensors.h
```diff
@@ -92,74 +93,85 @@ namespace OpenMEEG {
...
+                else if ( tokens.size() != ncol ) {
+                    std::cerr << tokens.size() << " != " << ncol << std::endl;
+                    std::cerr << "Problem while reading sensors file" << std::endl;
+                    std::cerr << "Each line should have the same number of elements" << std::endl;
+                    exit(1);
+                }
+                ++nlin;
+            }
+        }
+        if (nlin == 0) {
```
massich: not cover

OpenMEEG/src/EEGsensors.cpp
```diff
@@ -0,0 +1,127 @@
...
+        // init private members :
+        // positions
+        m_positions = mat.submat(0,nlin,0,3);
+        m_pointSensorIdx.resize(nlin);
+        // Sensor index
+        std::iota(m_pointSensorIdx.begin(), m_pointSensorIdx.end(), 0);
+        m_nb = nlin;
+    }
+
+    void EEGSensors::save(const char* filename) {
```
massich: not cover

OpenMEEG/src/EEGsensors.cpp
```diff
@@ -0,0 +1,127 @@
...
+                    m_labels.push_back(buf);
+                    labels.insert(buf);
+                }
+                Vector v(ncol);
+                iss >> v;
+                mat.setlin(i, v);
+                ++i;
+            }
+        }
+        if ( labeled and labels.size() != nlin) {
```
massich: not cover

OpenMEEG/src/EITsensors.cpp
```diff
@@ -0,0 +1,228 @@
...
+            m_radii.set(0.);
+        }
+        // find triangles on which to inject the currents and compute weights
+        findInjectionTriangles();
+        // Sensor index
+        std::iota(m_pointSensorIdx.begin(), m_pointSensorIdx.end(), 0);
+        m_nb = nlin;
+    }
+
+    void EITSensors::save(const char* filename) {
```
massich: not cover

OpenMEEG/src/EITsensors.cpp
```diff
@@ -0,0 +1,228 @@
...
+                ci_mesh_names.push_back(s_map);
+                ci_triangles.push_back(1);
+            }
+
+            triangles.push_back(current_nearest_triangle);
+            std::set<size_t> index_seen; // to avoid infinite looping
+            index_seen.insert(current_nearest_triangle.index());
+            if ( not almost_equal(m_radii(idx), 0.) ) {
+                // if the electrode is larger than the triangle, look for adjacent triangles
+                if ( current_nearest_triangle.area() < 4.*M_PI*std::pow(m_radii(idx),2) ) {
```
massich: not cover

OpenMEEG/src/EITsensors.cpp
```diff
@@ -0,0 +1,228 @@
...
+                }
+                m_weights(idx) = M_PI * std::pow(m_radii(idx),2) / triangles_area;
+            }
+            m_triangles.push_back(triangles);
+        }
+        for(size_t i=0;i<ci_mesh_names.size();++i)
+            std::cout<<ci_triangles[i]<<" points have been mapped to mesh "<<ci_mesh_names[i]<<std::endl;
+    }
+
+    SparseMatrix EITSensors::getWeightsMatrix() const {
```
massich: not cover

massich:
> I could use tests/test_sensors.cpp but this folder is not anymore used ?? 

Do we have tests that are not being run?

eolivi:
hey @massich 
actually I was waiting for someone to answer where should I write the test.
I repeat:
> I could use tests/test_sensors.cpp but this folder is not anymore used ?? @massich
The only running tests are in apps/ now, right ? does it cover non regression tests still ?

agramfort:
>
> hey @massich <https://github.com/massich>
> actually I was waiting for someone to answer where should I write the test.
> I repeat:
>
> I could use tests/test_sensors.cpp but this folder is not anymore used ??
> @massich <https://github.com/massich>
>
>
I guess it's a mistake if we forgot to migrate some tests :(

> The only running tests are in apps/ now, right ? does it cover non
> regression tests still ?
>
> it should

massich:
Well, this shows that we need a contribution guide. Better than https://github.com/openmeeg/openmeeg/blob/master/coding_guidelines.txt

agramfort:
the best way is to enforce this with a linter

we also need a what's new page we keep up to date along the way

>

massich:
What I meant is that our contributing guide is just a manual linter (which has limited utility) rather than being a contributing guide explaining how to contribute, run the tests, etc.

agramfort:
agreed

massich:
OpenMEEG uses no [C++ test framework](http://en.wikipedia.org/wiki/List_of_unit_testing_frameworks#C.2B.2B)
for TDD development such as [Google Test](http://code.google.com/p/googletest/)
or [catch2](https://github.com/catchorg/Catch2). Instead it writes fully fledged
programs that exercises the functionalities and subscribes them to `ctest` using
`CMakeLists.txt`. Which is equivalent but is less flexible and requires more
boilerplate.


In OpenMEEG we somehow do the distinction between unit-testing and integration
tests. In the current status there is no clear cut on what goes where, though.

we find unit-tests in two places:

```sh
$ find ./ -iname 'tests'             
./tests
./OpenMEEGMaths/tests
```

Here are two examples of how they get registered to the `ctest` runner.
[`load_geo.cpp`](https://github.com/openmeeg/openmeeg/blob/master/tests/load_geo.cpp#L7)
is registered here:

https://github.com/openmeeg/openmeeg/blob/master/tests/CMakeLists.txt#L16-L19

and [`full.cpp`](https://github.com/openmeeg/openmeeg/blob/master/OpenMEEGMaths/tests/full.cpp#L48) is registered here.
https://github.com/openmeeg/openmeeg/blob/05c7f0a8e79bd6760bea65d814dbb9ae00139fcb/OpenMEEGMaths/tests/CMakeLists.txt#L5

Both would generate their own binary and it would be executed. The former is a `main` that would
be executed through the shell with some parameters while the later is only a `main`.

Similarly we have integration tests in `./apps/`. Those use binaries that ~are~ (should be)
 distributed with the package (i.e. `om_geometry_info`) and exercise them against some data.
 Here is where the binary is created, exercised and distributed:

https://github.com/openmeeg/openmeeg/blob/05c7f0a8e79bd6760bea65d814dbb9ae00139fcb/apps/tools/CMakeLists.txt#L7

https://github.com/openmeeg/openmeeg/blob/05c7f0a8e79bd6760bea65d814dbb9ae00139fcb/apps/tools/CMakeLists.txt#L111-L114

https://github.com/openmeeg/openmeeg/blob/05c7f0a8e79bd6760bea65d814dbb9ae00139fcb/apps/tools/CMakeLists.txt#L116


To summarize I would write a little main inside some `$PATH/tests`. If possible
possibly stand alone main with no need for i/o, something that the compiler can
strip out most of it. Add it to the `CMakeLists.txt` to build and to test but not
to distribute. This should be enough.

At this point I would do it like the rest of tests and write the entire entry.
```
add_executable(my_test my_test.cpp)
target_link_libraries(my_test OpenMEEG) # or OpenMEEGMath whatever you have to link to but as little as possible

OPENMEEG_UNIT_TEST(my_test SOURCES my_test.cpp INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR} LIBRARIES OpenMEEGMaths)
```

By writing this piece I realized several things:
1 - Not all the tests are runed (and checking the git they have never been).
2 - there are  `apps/tools` that they are actually unit tests
3 - `./tests/` actually belong to `./OpenMEEG`
4 - using the `OPENMEEG_UNIT_TEST` macro we register the executable twice


cc: @eolivi

agramfort:
I restarted the appveyor build just to see if it helps

eolivi:
IMO this MR should be merged as well, since the travis failure is not due to this MR content (as in https://github.com/openmeeg/openmeeg/pull/368).

@papadop @agramfort @massich  ?

agramfort:
@eolivi this PR will now be hard to merge. I also agree that maybe it's a good idea to have separate classes per type of sensor. radii does not make sense for MEG, weights does not make sense for EEG

@papadop do you have an opinion on this?

agramfort:
@eolivi you are going to be in a merge conflict nightmare since the work of @papadop to have better python bindings...

larsoner:
The conflicts here are a bit tough. @eolivi I'll close, but feel free to reopen if you manage to rebase. At this point it might be easier to start from a fresh branch, though!

Issue #314, opened 2018-06-12
Write Non-regression test for realocable OpenMP necessary dll


massich:
https://github.com/openmeeg/openmeeg/blob/11d1db77dcb07107e17e747f8559324897926a89/CMakeLists.txt#L43

sarthakpati:
Curious how this issue will get solved.

massich:
I'm not sure how is the status. Do you have any ideas on how to solve it or want to give it a try?

sarthakpati:
#367 should fix this.

massich:
closing by #367. We still miss the regression test, but I think it is fine.

PR #367, opened 2019-05-17
fixing #314 using the same mechanism as https://github.com/CBICA/CaPTk/pull/490

sarthakpati:
Fixing https://github.com/openmeeg/openmeeg/issues/314

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,16 +39,3 @@ add_subdirectory(wrapping)
 # OpenMEEG Export
 ################
 include(cmake/Packaging.cmake)
-
-# XXX: Hot fix to be able to ship win binaries when OpenMP is active
-#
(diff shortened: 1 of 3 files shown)


agramfort:
@massich what do you think. At least CIs are happy

massich:
If that's the solution (as it seems to be) and the artifact binaries work
I'm +1. We should even remove the lines rather than comment them.

@cbica, can you show a braking case and how this pr fixes it? Thx

papadop:
And the comment above those lines or at least the part that is obsolete....

sarthakpati:
I can remove that comment (I kept it for consistency for you folks to begin with). So far, we haven't seen any breaking cases for this.

sarthakpati:
Turns out, that code was already there. I just removed the hard-coded path in the root CMakeLists file and uncommented that call.

agramfort:
Thx @sarthakpati

massich:
@sarthakpati <https://github.com/sarthakpati> Thx!!

sarthakpati:
You're welcome

PR #370, opened 2019-06-07
TST: add OpenMEEG/tests

massich:


massich:
WIP: the current status is that **compare_matrix** is not a target to make it depend on `OpenMEEG` target so that it finds where `options.h` actually are.

tests/CMakeLists.txt
```diff
@@ -1,248 +1,22 @@
...
-set(VALIDATION_EIT "${CMAKE_CURRENT_BINARY_DIR}/om_validationEIT")
-if (WIN32)
-    set(VALIDATION_EIT "${EXECUTABLE_OUTPUT_PATH}/om_validationEIT")
+# tests
+if (BUILD_TESTING)
+    OPENMEEG_TEST(check_load_geo
+        load_geo ${OpenMEEG_SOURCE_DIR}/data/Head1/Head1.geom ${OpenMEEG_SOURCE_DIR}/data/Head1/Head1.cond)
+    OPENMEEG_TEST(check_test_mesh_ios
+        test_mesh_ios ${OpenMEEG_SOURCE_DIR}/data/Head1/Head1.tri)
 endif()
```
agramfort: this just moved the tests and it broke appveyor. These tests were passing and now they don't? https://ci.appveyor.com/project/openmeegci/openmeeg/builds/25139666

I don't get the logic here. The diff is huge as you move the majority of the code to a different file.

agramfort: I am starting to understand the problem :)

massich:
what I don't understand is that we are adding 300+ new tests and the coverage increases less than `5%`

agramfort:
if it works can you put back the test to one file rather than having them in other.cmake? it hides the content of the PR and I don't think it's necessary

massich:
I can refactor the cmake in a different PR. yes

tests/CMakeLists.txt
```diff
@@ -246,3 +255,5 @@ foreach(DIP 1 2 3 4 5 6)
         endforeach()
     endforeach()
 endforeach()
+
+include(clean_tests.cmake)
```
agramfort: what is this file?

massich: its a test that 'cleans' the comparision tests. It was in the original `CMakeLists.txt` I'm not sure if I would end up keeping it or not

PR #371, opened 2019-06-12
add apps/presmooth as a target

massich:
This PR adds back `presmooth` as a target. Which is impossible that had compiled since #11. Where the following change happen (see [this](https://github.com/openmeeg/openmeeg/pull/11/files#r292943572) and [this](https://github.com/openmeeg/openmeeg/pull/371#discussion_r292973764)):

```diff
- #include "mesh3.h"
+ #include <mesh.h>
```
This change made `SparseMatrix SmoothMatrix = SourceMesh.gradient();` and `const Vector &AiVector = SourceMesh.areas();` impossible to compile, so we were not compiling it.

I've checked what was distributed as a binary and the last time `presmooth` was there was 2.2:

```sh
/tmp
(base) ❯ find OpenMEEG-2.* -type f -executable | grep smooth
OpenMEEG-2.2.0-Linux64.amd64-gcc-4.4.1-OpenMP-static/bin/om_mesh_smooth
OpenMEEG-2.2.0-Linux64.amd64-gcc-4.4.1-OpenMP-static/bin/om_presmooth
OpenMEEG-2.3.0.1-MacOSX-Intel-static/usr/bin/om_mesh_smooth
OpenMEEG-2.3.0.1-MacOSX-Intel-static/Users/travis/build/openmeeg/openmeeg/build/OpenMEEG/install/bin/om_mesh_smooth
OpenMEEG-2.4.1-Linux/bin/om_mesh_smooth
```

The alternative is to merge #372

### DIFF
--- a/apps/presmooth.cpp
+++ b/apps/presmooth.cpp
@@ -64,7 +64,7 @@ int main(int argc, char **argv) // TODO a quoi ça sert ?
 
     SourceMesh.load(argv[1]);
 
-    SparseMatrix SmoothMatrix = SourceMesh.gradient();
+    SparseMatrix SmoothMatrix = SourceMesh.gradient_norm2();
 
     // write output variables
     SmoothMatrix.save(argv[2]);
--- a/OpenMEEG/src/mesh.cpp
+++ b/OpenMEEG/src/mesh.cpp
@@ -49,7 +49,7 @@ namespace OpenMEEG {
         if (m.allocate_) {
             allocate_     = true;
             all_vertices_ = new Vertices; // allocates space for the vertices and copy
-            all_vertices_->reserve(m.nb_vertices()); 
+            all_vertices_->reserve(m.nb_vertices());
             std::map<const Vertex *, Vertex *> map; // for the triangles
             for (Mesh::const_vertex_iterator vit = m.vertex_begin(); vit != m.vertex_end(); ++vit) {
                 add_vertex(**vit);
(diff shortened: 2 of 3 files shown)


apps/presmooth.cpp
```diff
@@ -64,7 +64,7 @@ int main(int argc, char **argv) // TODO a quoi ça sert ?
 
     SourceMesh.load(argv[1]);
 
-    SparseMatrix SmoothMatrix = SourceMesh.gradient();
+    SparseMatrix SmoothMatrix = SourceMesh.gradient_norm2();
```
massich: Signature from `mesh3.h` gradient and `mesh.h` `gradient_norm2` is not the same. 

See when `gradient` changed to `gradient_norm` before becoming `gradient_norm2`:
```sh
commit 21c7424879629a07a5b86c9dac43509571c62710
Author: Emmanuel Olivi <emmanuel.olivi@gmail.com>
Date:   Wed Dec 18 18:06:01 2013 +0100

    Remove a lot of annoying warnings and add a gradient norm function.

diff --git a/src/mesh.h b/src/mesh.h
--- a/src/mesh.h
+++ b/src/mesh.h
@@ -172,3 +174,3 @@
         /// \brief Compute the surfacic gradient
-        void gradient(SparseMatrix &A) const;
+        void gradient_norm(SymMatrix &A, double coeff) const;
```

massich:
ping: @agramfort , @papadop , @eolivi 

Would you resurrect `presmooth` or shall we remove it altogether? (I'm +1 to the later)

PR #372, opened 2019-06-12
remove presmooth

massich:
closes #371 
This PR removes `presmooth` without touching `data/*/Makefile[example.win]` which is using it. see this as example:

https://github.com/openmeeg/openmeeg/blob/d7d03eef4faa6b6a267b5dc30b7893eb974382c3/data/Head1/Makefile#L170-L171


agramfort:
thx @massich

PR #375, opened 2019-06-14
MRG: Move compare_matrix to apps/tools

massich:
requires #374 to have smaller diff.


agramfort:
you need to rebase.

can you try to remove dep on OpenMEEG::OpenMEEG ?

massich:
`compare_matrix.cpp` uses `"options.h" to handle the arguments. This belongs to `OpenMEEG::OpenMEEG`

apps/tools/CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ target_link_libraries(om_check_geom OpenMEEG ${VTK_LIBRARIES})
 add_executable(om_mesh_to_dip mesh_to_dip.cpp)
 target_link_libraries(om_mesh_to_dip OpenMEEG ${VTK_LIBRARIES})
 
+add_executable(compare_matrix compare_matrix.cpp)
```
agramfort: maybe we call it test_compare_matrix to be consistent? I guess it's not installed/shipped anyway so we can rename 

I would do the same for test_load_geo

massich: yes we have several of them. We can change them all at once in a separated PR.

I guess that `validateEIT` also falls in this category

agramfort:
green. Merging

PR #376, opened 2019-06-14
WIP: remove makefiles

massich:
The makefiles do no longer work, and they don't apply. 

I assumed that all the cases are cover by some test in cmake but did not double check it.


agramfort:
@mclerc @papadop I agree to merge this.

If you think this looses some information for users then it should be moved to the documentation or https://github.com/openmeeg/openmeeg_sample_data

papadop:
I did not even new that those were existing !! +1 for me. If the idea is to provide a processing pipeline, it should be in a tools directory and generic, not in each of the HeadX directories...
If it can be made generic in some way, it might be a nice way to run the full pipeline on user provided testcases (but I have not analyzed if it is possible).

agramfort:
cool. thx @massich

PR #381, opened 2019-09-10
WIP: Develop

j-l-s:
Just to follow the actual development. Do not pull into master repo.


massich:
why can't I push in this branch?

wrapping/src/openmeeg_path_setter.py
```diff
@@ -0,0 +1,26 @@
...
+        omdir = os.path.join( topdir, l)
+        sys.path.append(d)
+        found = 1
+        break
+    l = os.path.join(l , "..")
+
+if found == 0:
+    print("Cannot find locally compiled _openmeeg.so file")
+    print("Please goto openmeeg repository or a subdir of it")
+    exit(-1)
```
massich: `sys.exit(-1)` or `from sys import exit`

wrapping/src/openmeeg_path_setter.py
```diff
@@ -0,0 +1,26 @@
...
+# find openmeeg lib
+#
+import sys
+import os
+
+# import locally built openmeeg
+topdir = os.getcwd()
+omdir = ""
+
+# heuristic
```
massich: why are we doing this adhoc path updating?

papadop:
Yes. Me niether. Jean-Luc can you allow us to push commits on this clone ???

massich:
I've open a PR updating travis. @j-l-s it would be great if you could merge it.

papadop:
@massich you should have received an invitation.

wrapping/src/openmeeg_path_setter.py
j-l-s: Sorry, this is my mistake. This heuristic allows us to run the python script from anywhere inside the OpenMeeg topdir without any extra environment variable... (the scirpt basically find the openmeeg topdir, and then finds the data and the _openmeeg.so file). I didn't check it properly then running the test. I guess I just broke the python tests.

This script should be removed to get the old behavior back.

wrapping/src/openmeeg_path_setter.py
```diff
@@ -0,0 +1,26 @@
...
+found = 0;
```
massich: `;`

massich:
This should be good to go, and we can keep iterating in subsequent PRs. 
We just need to make sure that the coverage issues reported by codecov wont be a problem.

wrapping/src/test_matrix.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+
+import openmeeg_path_setter as ps
```
agramfort: do we need this hack to make the CIs happy?

massich:
They wont be green. I'm fixing some stuff needed in order to remove the path hack.

wrapping/src/CMakeLists.txt
```diff
@@ -72,13 +73,26 @@ install(FILES  ${OpenMEEG_BINARY_DIR}/wrapping/src/openmeeg.py
...
+  "test_vector.py"
+  )
+
 # Test Python wrapping
 #
 # XXXX: we copy all python scripts from source to binary so that openmeeg.py is in the path. There should be a manner of just adding
 #       it to the python path for the testing environment and allow to execute the python scripts from source.
 if (BUILD_TESTING)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E copy ${OpenMEEG_SOURCE_DIR}/wrapping/src/test_python.py ${OpenMEEG_BINARY_DIR}/wrapping/src/)
-    add_test(openmeeg_python ${PYTHON_EXECUTABLE} ${OpenMEEG_BINARY_DIR}/wrapping/src/test_python.py --path ${OpenMEEG_SOURCE_DIR}/data)
```
massich: we modified `openmeeg_python` to `openmeeg_python_XXX` without porting any of the properties. That were previously set.

massich:
Thx to everyone!

agramfort:
Thanks @massich !

PR #382, opened 2019-10-02
MNT: update travis to bionic

massich:
This PR:
- would fix travis (for second entry of our matrix)
- update linux virtual machines to use ubuntu-1804:bionic


massich:
The build has been replicated https://github.com/massich/openmeeg_builds/compare/fix_travis using docker the weird thing is that its working there and not in travis.

PR #385, opened 2019-10-11
Docstrings

papadop:
A small changes to re-enable docstrings in the python module...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
         - ENABLE_PYTHON=ON
         - ANALYSE=ON
         - ENABLE_COVERAGE=ON
-        - BUILD_DOCUMENTATION=OFF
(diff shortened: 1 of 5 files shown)


papadop:
PS: I leave the work of adding a travis/appveyor config that checks that documentation is built.
I can do this but I'm unsure to which build you prefer to add it.
Obviously, with the current checker, the only important modification (in openmeeg.i) of this PR is not tested....

papadop:
Is travis build with python2 or python3 (to understand why the docstring.i file has not been generated).

papadop:
OK it is probably related to a parallel build and a missing dependency. I pushed a temptative fix.

massich:
Everything is build in python 3.

Python 2 is no longer supported on 2020

massich:
We still build it in openmeeg-feedstock until then, though

PR #388, opened 2019-10-24
WIP: More Python bindings

agramfort:


papadop:
Let's see what testers will say. 
**_For the first time, we are able to import geometries directly from python..._** 
But this is only tested very lightly by checking the info() of the geometry.

The syntax is:
# Assuming cortex, skull and scalp are meshes imported with the command:
# mesh = om.Mesh(Vs,Ts) where Vs is a numpy array of vertices and Ts is a numpy array of 
# triangles (see test_mesh.py for details).
# See test_make_geometry.py for a complete example...

interfaces = {
    "interface1": [(cortex,om.OrientedMesh.Normal)],
    "interface2": [(skull,om.OrientedMesh.Normal)],
    "interface3": [(scalp,om.OrientedMesh.Normal)]
}

domains = {
    "Scalp" : ([('interface2',om.SimpleDomain.Outside), ('interface3',om.SimpleDomain.Inside)],1.0) ,
    "Brain" : ([('interface1',om.SimpleDomain.Inside)],1.0),
    "Air"   : ([('interface3',om.SimpleDomain.Outside)],0.0),
    "Skull" : ([('interface2',om.SimpleDomain.Inside), ('interface1',om.SimpleDomain.Outside)],0.0125)
}

geometry  = om.make_geometry(interfaces,domains)

For the time being, make_geometry (actually OpenMEEG) inconveniently prints all the geometry info... That needs to be changed, I guess...

papadop:
Can someone reproduce the Mac problem with python enabled. I cannot understand why the mesh type would be hashable in linux and not in MacOS. It seems to be a pure python problem (we add an object to a set) even though the object itself is an openmeeg object (here a mesh).

papadop:
Strangely linux and MacOS use the same python version 3.6.

agramfort:
Can you rebase after CI fix ?

papadop:
See PR #395

PR #389, opened 2019-10-25
Mesh work (WIP)

papadop:
During the python coding sprint, I have has a look at the Mesh classes.
I do not like them much (it's not new)...
Yet, I think we can improve things a little.
This is a first step which removes all the logic allowing to copy meshes. The copy constructor is deleted, and we use mesh references whenever copies were made. In some cases, I had to change slightly the logic, but I think that the initial intent of the code is respected in those cases.
One minor point is that swig does not know to wrap vectors of classes with no copy constructor (here the class Meshes==std::vector<Mesh>). Until I find a workaround (there seems to be one), I disabled the wrapping of this class. But I'm not totally sure that we really need it either....
There are two methods that can be removed (not yet done. They are only #if'ed out for now. see mesh.cpp).
This is a POC/WIP. Let us see what tester will say....

I intend to go a little further in the restructuring of this code (the all_vertices stuff seems in particular weirdly crafted into the mesh classes).

### DIFF
--- a/wrapping/src/openmeeg.i
+++ b/wrapping/src/openmeeg.i
@@ -111,7 +111,7 @@ namespace std {
     %template(vector_vertex) vector<OpenMEEG::Vertex>;
     %template(vector_pvertex) vector<OpenMEEG::Vertex *>;
     %template(vector_triangle) vector<OpenMEEG::Triangle>;
-    %template(vector_mesh) vector<OpenMEEG::Mesh>;
+#    %template(vector_mesh) vector<OpenMEEG::Mesh>;
     %template(vector_string) vector<std::string>;
     %template(vector_interface) vector<OpenMEEG::Interface>;
 }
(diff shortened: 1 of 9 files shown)


wrapping/src/openmeeg.i
```diff
@@ -111,7 +111,7 @@ namespace std {
     %template(vector_vertex) vector<OpenMEEG::Vertex>;
     %template(vector_pvertex) vector<OpenMEEG::Vertex *>;
     %template(vector_triangle) vector<OpenMEEG::Triangle>;
-    %template(vector_mesh) vector<OpenMEEG::Mesh>;
+#    %template(vector_mesh) vector<OpenMEEG::Mesh>;
```
agramfort: cleanup? what did you have to change this?

papadop: As I explained, swig does not know to wrap std::vector<X> when X has a deleted copy constructor. There are workarounds I need to investigate. But on the other end, I'm not sure we really need the Meshes. At least for now this was not tested at all....
The travis failure shows that one of the methods I removed was useful in the vtk load.
I will eventually restore that one at least to see if this solves the travis pb.

papadop:
This is no longer needed except maybe for the vector<Mesh> in openmeeg.i for which I will try a different approach in coming PR.

larsoner:
Okay for me to "delete branch" then? It would be nice to clean up any stale branches like this from `openmeeg` !

PR #390, opened 2019-10-25
Correct vertex access for triangles.

papadop:
Some ugly things I noticed a while ago but never corrected.
Triangle edges were iterated over by going through vertices (i,i+1) for i from 0 to 2.
For i=2, this means that we were accessing to the fourth vertex of a triangle. s this is wrong this was handled by having a modulo 3  in the vertex accessor.
This PR corrects the problem at its source by never accessing the "fourth vertex".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.9)
 project (OpenMEEG VERSION 2.4.9999 LANGUAGES C CXX)
 
-set(CMAKE_CXX_STANDARD 11)  # use c++11
+set(CMAKE_CXX_STANDARD 17)  # use c++17
 
 set(OpenMEEG_HEADER_INSTALLDIR include) # XXX I don't know why this should be here
 
--- a/OpenMEEG/include/vect3.h
+++ b/OpenMEEG/include/vect3.h
@@ -53,11 +53,9 @@ knowledge of the CeCILL-B license and that you accept its terms.
 
 namespace OpenMEEG {
 
-    /** \brief  Vect3
+    inline double sqr(const double& x) { return x*x; }
 
-        Mesh Class
-
-    **/
+    /// \brief  Vect3
 
     class OPENMEEG_EXPORT Vect3 {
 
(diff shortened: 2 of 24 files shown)


papadop:
The second commit slightly improves the == test.

agramfort:
@papadop CIs are not happy. Please have a look before doing more

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.9)
 project (OpenMEEG VERSION 2.4.9999 LANGUAGES C CXX)
 
-set(CMAKE_CXX_STANDARD 11)  # use c++11
+set(CMAKE_CXX_STANDARD 17)  # use c++17
```
agramfort: was this really necessary? does an ubuntu or debian stable support this?

OpenMEEG/include/vect3.h
```diff
@@ -53,102 +53,101 @@ knowledge of the CeCILL-B license and that you accept its terms.
...
+        // De Munck : Good sign directly
+        const Vect3& V0 = *this;
+        const Vect3& Y1 = V1-V0;
+        const Vect3& Y2 = V2-V0;
+        const Vect3& Y3 = V3-V0;
+        const double y1 = Y1.norm();
+        const double y2 = Y2.norm();
+        const double y3 = Y3.norm();
+        const double d = det(Y1,Y2,Y3);
+        return (fabs(d)<1e-10) ? 0.0 : 2*atan2(d,(y1*y2*y3+y1*dotprod(Y2,Y3)+y2*dotprod(Y3,Y1)+y3*dotprod(Y1,Y2)));
```
agramfort: this code after the atan2 is duplicated in multiple places.

papadop:
Unfortunately in this case it's not only cosmetic.... It involves homogeneizing the access to vertices of triangles, and generally helps making the code clearer using the new range-for loops (which are better handled by OpenMP than the pointer based loops).

For me the new writing is much clearer (it's not totally finished)....
There are basically 5 types of changes:
1) the homogeneization of t[],t(), s[123]() and vertex() into one single method (vertex()).
    This also removes index of vertices using i%3  and some logical errors in the code (see edges 
    below).
2) the conversion of the numerous 
    for (whetever_unclear_iterator [vmt]it=sthg.begin();[vmt]it!=sthg.end();++[vmt]it) into
    for ([const] auto[&] {vertex,nesh,triangle} : container)
    which generally make the code clearer. This is not yet totally finished.
3) Introducing the notion of edges, which clarifies some pieces of code and makes it more correct.
4) A few other uses of modern C++ features (construction calling other constructors,constexpr,...)
5) Removing some unecessary functions which were called only once and for which I cannot see a good reason (assembleSensor and assenbleSources).
  
In the end it is a lot of very similar changes...
The question is "are we confident in the testers or not"....

agramfort:
>
> In the end it is a lot of very similar changes...
> The question is "are we confident in the testers or not"....
>
I can only testify for the nested geometries. The non-nested code was
merged without review so I cannot guarantee it's tested properly.
First thing to see would be to see how good is the coverage.
https://codecov.io/gh/openmeeg/openmeeg
If you modify code that is not covered you can be sure
that there is no safe net..

CMakeLists.txt
papadop: This is not fully necessary. I can work around its need (it is just that static constexpr unsigned x[]; changed semantics between c++11 and c++17. If we need to stick with c++11, I need to add a new file to define the static member....
I think the support is there for quite some time.... As far as I can tell c++17 support started from g++5 and was mostly complete by g++-7, That does not mean that there are no bugs. I surely do not yet totally understand the ubuntu ci problem.... It fails on a simple c++11 feature.... I need to investigate more.

papadop: We are currently at gcc-9 and soon gcc-10...

papadop:
There are test cases for the non-nested geometries (see the NN tests).
I'm slightly more worried for the side stuff such as mesh correction (orientation etc) in which most of the ugliest stuff was.

papadop:
And I forgot a change:
6) Replaced V1*V2, V1^V2 and V.det(V1,V2) by dotprod(V1,V2),crossprod(V1,V2) and det(V,V1,V2).
    Also renamed solangle into solid_angle and flip() by change_orientation.
    I think this makes a lot of code much much clearer.
    The conversion is not yet totally complete.... I have not removed the ^ operator I think, so there
    might still be other uses...

agramfort:
I will try to look this week when Travis is green

papadop:
These changes are already integrated in main.

PR #392, opened 2019-11-01
clang-format all code

eolivi:
it will pollute our git history, but we should have done this before ....
What do you think ? @papadop @agramfort @massich ?
If Yes, we should also enforce new PR to fulfil this format...

` find OpenMEEG/ OpenMEEGMaths apps tests -name "*cpp" -or -name "*.h" | xargs clang-format -i `

feel free to vote, accept or reject this PR


papadop:
We were discussing exactly the need to add something  like that with @agramfort  last week. I did not know clang had this capability. This is an interesting  track...

Just two things:
- we are on the verge on committing quite many changes, so the timing is not good as this will complicate a lot the merges....  We should wait for these big merges to be done. By the way I have several questions for you about some very similar code and about isolated, current barriers and several other things....
- I have had a very quick look at one single file, I'm not always too happy with the style (namespace not indented, indentations of 2 spaces, etc). Is this configurable ??

A last thing is that would require everyone to have clang installed. This is probably much better than many other tools as it is becoming quite common.... Does github provide a facility like this, in which case this will be doable on the server side ?

agramfort:
Lets do this after current PRs on python wrapping are done. 

Yes we can configure this. See

https://github.com/InsightSoftwareConsortium/ITK/blob/master/.clang-format

For an example

I would not add this as commit book or anything that requires to have clang installed but we should add this to one Travis build

eolivi:
@papadop I am sure, you are doing a good job on your PR on meshes, and that these classes need to be refactored.
But as @agramfort pointed out, it is not easy to review because of the format.

IMO we should go soon for clang-format, (let's not discuss how many spaces, and so on... rrrrrr let clang-format do).

I can answer on some questions for meshes, but since all the work on isolated or current barriers, happen later than me I don't know this part.

I remember that the tricky part for non nested geometries was not to have duplicated points (e.g one border described by a mesh A, and the same border described in mesh B), that was the point of all_vertices ...

massich:
I'm super +1 in adding something that allow for consistency in the coding style.

massich:
The way to do those things usually is to use a tool with a config that we more or less like. Apply the command and commit the result.

massich:
The way to do those things usually is to use a tool with a config that we more or less like. Apply the command and commit the result.

eolivi:
@agramfort no idea what the fails on MacOs are about ...

agramfort:
master is green on mac os?

i would start from there

larsoner:
I'll close this but agreed once any useful PRs are merged CIs should be updated to ensure code conforms to clang-format

PR #393, opened 2019-11-15
Repare appveyor

papadop:
Attempt to repare appveyor by downgrading conda (as seen in a web page, reference to be added later).


agramfort:
can you try to do it after branching from current master? it will not interact then with the recent changes you made. Just to decouple potential issues.

papadop:
Yes. Mistake on my part. I tried to do a quick fix just before a meeting...

PR #394, opened 2019-11-15
Attempt to fix conda failure in appveyor

papadop:
On internet two solution are proposed:
- downgrade to conda 4.3.34
- avoid path manipulation and instead call a script (see https://github.com/conda/conda/issues/8836)

Attempting the last one.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -58,7 +58,7 @@ install:
   ############################################################################
   - "git clone --depth 1 git://github.com/astropy/ci-helpers.git"
   - "powershell ci-helpers/appveyor/install-miniconda.ps1"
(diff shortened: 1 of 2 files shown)


papadop:
Seems appveyor is repaired...
Why travis fails is a nonsense and probably just a temporary network problem.

papadop:
My last comment means I think we can merge.... And then evaluate the python branch state on windows... I'm a little bit afarid as I have seen that we use VS 2015 and I wonder whether it has the new C++ features we use (or we will have to upgrade the compiler).

agramfort:
Perfect thanks @papadop

massich:
great Job!

PR #395, opened 2019-11-16
New PR for python bindings

papadop:
This is a continuation of PR #388 which was closed (because the develop branch was moved and replaced by a new - cleaner ? - one). This was needed to rebase to the upstream master. I think that the problem is that when the develop branch was merged after the first code sprint the changes were squashed. Then the patches were incorporated twice (once from the original branch and once during the merge with the squashed master) and then merged. So I had two choices:
- Merge again (but this will have to be repeated in any newer rebase).
- Reapply the newer patches starting from the current master. This is what I did....


papadop:
appveyor is happy !!
Remains the MacOs python pb....

agramfort:
I tried to build this branch on my mac using my local setup which is using brew for swig / cmake,
the system for clang compiler and anaconda for python and it worked fine.

maybe use clang and not gcc on the mac build?

papadop:
To me it seems more a python related problem than a compiler one. The message in the failing test is a python message, I think... Very unclear to me though....

agramfort:
can someone replicate it around you?

>

papadop:
I do not have Macs (or very very outdated, which will make compiling OpenMEEG difficult). I will see if Jean-Luc or Julien can replicate, but they usually use the very latest OS so they might have trouble replicating either. Not sure to know when they will have time.

papadop:
OK. Jean-Luc replicates the problem. But I still do not understand well what happens. I thought it might be python-related (he uses 3.7.4 and I have 3.7.5), but I recompiled the same python as he has, and the problem does not happen. He also uses swig 4.0.1 where I have 3.0.12/ I will compare the generated code, but since there are so many lines I'm not sur I will get sthg. I will try to get a more recent version of swig...

Alex can you provide the python and swig versions of your environment ?

papadop:
Final remark: I do not understand the unhashable error. The swig generated mesh class has both the __eq__ and __hash__ methods in both cases (ie my working context and Jean-Luc non-working context). So the comments on the web relating the unhashability to the presence of the __eq__ method does not seem to be relevant (or I need to understand how they are relevant).

papadop:
OK. I understood.
This problem appears with swig 4.0.1 which no longer defines the __hash__ method....
Try this code:
```
import openmeeg as om
m = om.Mesh()
print(m.__hash__)
```
With swig 3.0.12 it gives:
<method-wrapper '__hash__' of Mesh object at 0x7f794f31f990>
With swig 4.0.1, it gives:
None


Now, I need to find the proper fix...

agramfort:
I propose to merge this one. Than https://github.com/openmeeg/openmeeg/pull/396 and then run the full code formatter adding this to the CI to avoid going back and forth with formatting issues.

ok for you @papadop @eolivi ?

papadop:
We can merge this, even though it is not perfect (there are still pbs with reference counting which spurred the #396 work. Yet the problems are no worse than what they were, and the branch brings several improvements notably with the wrapping of geometries.

papadop:
As for the formatting, I just have played with clang-format and it seems to be still evolving so that there are some formatting options with clang-10 which are not yet available with the clang version I have on my brand-new fedora 31 (and I'd expect that it will be even worse with tha github/travis/appeyor machines). I'm afraid that if we change now, we will have to change again later....

I have not re-looked at more mature tools (uncrustify was the one I chosed last time I did such a work), but I'm afraid that those might not be as available as clang-format,

papadop:
There are actually two different things.
1) the actual formatting of the code.
2) what appears in diffs.

For the second there are some git options that might be relevant (there are several white-space/blank lines options):
https://git-scm.com/docs/diff-options/2.5.1

For the first, whatever I have seen remains unsatisfactory.... but here are somenice resources I found:
https://medium.com/millennial-falcon-technology/reformatting-your-code-base-using-prettier-or-eslint-without-destroying-git-history-35052f3d853e
https://stackoverflow.com/questions/48124523/using-git-how-to-auto-format-when-pull-clone-fetching-from-github
https://medium.com/@elliotchance/reformatting-your-codebase-with-git-filter-branch-21feded7c423
https://dev.to/sandordargo/how-to-automatically-format-your-c-code-regardless-the-ide-you-use-2pm2

agramfort:
ok merging. thanks @papadop @j-l-s 

I'll start an mne-openmeeg project to obtain forwards from mne objects in the next couple of days

agramfort:
I would not aim at perfection. let's just adopt something commonly accepted.
We have to find a compromise.

>

PR #396, opened 2019-12-05
Use std::shared_ptr instead of our own RC* function.

papadop:
This PR removes the use of our own shared pointer solution in favor of the standard std::shared_ptr. Hopefully it will also help to improve the python binding as the std::shared_ptr is already handled by swig. This also remove some files which we no longer have to maintain and simplifies a little bit the code. Tested on fedora 31, let's see what the other platforms will say....


papadop:
Replaced by PR #399 .

PR #397, opened 2019-12-08
reveal bug with make_geometry

agramfort:
@papadop here is just a tiny diff to show a bug I observed while trying to obtain forwards using MNE

you'll see that the make_geometry function does not produce the same type of Geometry as when going file based. It does not say it's nested.

can you have a look?


papadop:
OK. I know what is happening. There is some wrong duplication of meshes.... In make-geometry, we import meshes after creating the domains. This is certainly the reason....

agramfort:
can you have a look?

I am pushing stuff here https://github.com/mne-tools/mne-openmeeg

>

papadop:
Superseded by PR #410.

PR #398, opened 2019-12-08
fix some pep8 errors

agramfort:
@papadop this is just a cosmestic PR to format the python more following the pep8 standard.


papadop:
Not  a big fan of white lines insert between imports, but let's merge anyway.
Do you use a tool to do that ?

agramfort:
Yes https://pypi.org/project/flake8/

PR #399, opened 2019-12-14
Use std::shared_ptr instead of our own RC* function.

papadop:
Replaces PR #396 (that I have not been able to rebase)....
This PR removes the use of our own shared pointer solution in favor of the standard std::shared_ptr. Hopefully it will also help to improve the python binding as the std::shared_ptr is already handled by swig. This also remove some files which we no longer have to maintain and simplifies a little bit the code. Tested on fedora 31, let's see what the other platforms will say....


papadop:
It looks like this is an heisenbug in travis. 2 failures previously (linux), one now (mac which was passing previously, and the change is a dummy one - removal of a comment -). So I think this is ready to be merged...

Please do not squash the merge, as it makes further merges difficult (this was done with the previous PR, which obliged me to do a manual merge with all the time and the risks this incurs).

agramfort:
@papadop you've been missing my comments or just ignoring them but I am not ok with adding a dependency on boost and I don't want to merge code that makes CIs red.

sorry for being opinionated. I find this other PR https://github.com/openmeeg/openmeeg/pull/397 more problematic and urgent as it reveals an actual bug that prevents me from moving forward with mne-openmeeg integration

papadop:
Sorry but I have not seen your comment... not on this PR at least.
The depedency on boost is only a compile time dependency and only for windows, which does not have this standard header, so in my view it is not a problem from the user point of view. Note also that boost is pre-installed on appveyor, so that I did not even had to change the .yml...

As for the red, read my comment. It is a problem with travis, not with the code. Two successive runs (with no change) result in two different results... If travis is not reliable, then we will have to re-consider this policy.... Hopefully it is just a temporary problem.

papadop:
As for PR #397, I'm on it... But the solution is more complex than you may think. The problem is hidden in the mess of our mesh/geometry structure and a probable solution is .... shared_pointer !!!
PR 397 is because the vertices obtained from the python construction are duplicated... So the nested domains are not detected properly. I think I explained this in the PR. 

I spent most of my week-end in trying to clean the mesh/geometry code to the point it will be more manageable.... Yet another can of worms I opened !!! So I'm on it, but it evolves as fast as I have time to deal with it.

papadop:
Travis is green ;-) 
Thank's for the tip allowing to restart only the failed job !!!

cmake/utils/CheckCXXFeatures.cmake
```diff
@@ -85,31 +85,32 @@ macro(CHECK_CXX_STANDARD_LIBRARY)
...
-    # CHECK_CXX_FEATURE(HAVE_VALARRAY                      have_valarray.cpp                 "has valarray<T>")
-    # CHECK_CXX_FEATURE(HAVE_COMPLEX_MATH1                 have_complex_math1.cpp            "has complex math functions")
-    # CHECK_CXX_FEATURE(HAVE_COMPLEX_MATH2                 have_complex_math2.cpp            "has more complex math functions")
-    # CHECK_CXX_FEATURE(HAVE_IEEE_MATH                     have_ieee_math.cpp                "supports IEEE math library")
-    # CHECK_CXX_FEATURE(HAVE_SYSTEM_V_MATH                 have_system_v_math.cpp            "supports System V math library")
-    # CHECK_CXX_FEATURE(HAVE_MATH_FN_IN_NAMESPACE_STD      math_fn_in_namespace_std.cpp      "has C math functions in namespace std")
-    # CHECK_CXX_FEATURE(HAVE_MATH_ABSINT_IN_NAMESPACE_STD  math_absint_in_namespace_std.cpp  "has C math abs(integer type) in namespace std")
-    # CHECK_CXX_FEATURE(HAVE_COMPLEX_MATH_IN_NAMESPACE_STD complex_math_in_namespace_std.cpp "supports complex math functions are in namespace std")
-    # CHECK_CXX_FEATURE(HAVE_ISNAN_IN_NAMESPACE_STD        isnan_in_namespace_std.cpp        "has isnan function in namespace std")
```
agramfort: FYI @papadop  these had been commented out as it was not used and it was increasing cmake bulid time for nothing.

papadop: Yes pre-approved. Maybe you can integrate it with your old CMake removal. I already have merged this branch....

agramfort: see https://github.com/openmeeg/openmeeg/pull/403

PR #401, opened 2019-12-23
Check OpenMP support in cmake.

papadop:
With these changes, OpenMP support is checked in cmake (instead of trying to deduce OpenMP support from compiler macros).

Alex can you check if this corrects your catalina build (it should unless I made mistakes).


agramfort:
with the last commit it works @papadop 

it would be great if we don't open PRs from the upstream repo but using our clone. Otherwise it pollutes the full repo and people doing a git clone from here will get all the mess of historical branches.

papadop:
OK for working from my clone... although I do not believe there are that many devs.... It seems that PR from different repos have less checking (no branch and push variants in travis and appveyor).
For the change, I'm OK with it, but I do not understand how this can change from failure to pass...

papadop:
The advantage of being in the main repo is that other can also make some commits..... ;-)

agramfort:
One can also push to a PR from a different fork exactly to avoid having dead branches in upstream repo

PR #403, opened 2019-12-26
comment unnecessary CXX checks

agramfort:


papadop:
Sorry, I do not see the point. Looking a little bit better than previously (I was thinking you were suppressing already commented macros). The macro you modify is not called in OpenMEEG. We could suppress completely the macrp, but as the code comes from another project, I'd prefer to keep it "as is" as it simplifies synchronizing changes....

Some cleanup is necessary in this set of macros as cmake now does its own C++ feature check. But we need to understand what it does and keep the complementary tests.

agramfort:
well the point is that we don't make use of these. Someone reading the code
will start to wonder
why it's there (Sik for example...). The fact that it's because you use it
in other projects is not
obvious to new comers. I would like to reduce the barrier to start
contributing to openmeeg.
This is one step in this direction.

>

papadop:
For what it's worth. Out of all these tests, we only use CHECK_CXX_OPENMP_SUPPORT, which itself only checks that we have ab OPENMP with the capabilities of OpenMP 5.0 (range for loops, iterator loops and unsigned loops). So I guess, we can replace this with a version check with find_package(OPENMP).... But is it OK to loose support for older compiler with OpenMP version < 5.0 ? It was released  in Nov. 2018... Checking more thoroughly, it seems that VC++ is lagging behind !!!

Otherwise it will also be a nice cleanup in the code which has a lot of alternatives such as:
#if defined NO_OPENMP || defined OPENMP_RANGEFOR
#elif defined OPENMP_ITERATOR

larsoner:
> So I guess, we can replace this with a version check with find_package(OPENMP)

+1 for removing all other checks. Then we can say we just use c++17 plus (optionally) OpenMP

> Checking more thoroughly, it seems that VC++ is lagging behind !!!

Unfortunately we need Windows support, and we need it to work on Python. I'm not sure how good the support is for OpenMP + Windows nowadays, but I'm guessing "not good".

That being said, there is this infrastructure in conda-forge:

https://github.com/conda-forge/openmeeg-feedstock/blob/main/recipe/meta.yaml

The latest Windows build there did seem to have / use OpenMP:

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=541794&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=66e4a587-a723-53e9-873c-6c9a2c5971ea&l=527

Looking at the latest wheel build, it also seems to have had it:

https://github.com/openmeeg/openmeeg/runs/7597415119?check_suite_focus=true#step:4:476

So macOS might actually be the issue here:

https://github.com/openmeeg/openmeeg/runs/7597415095?check_suite_focus=true#step:4:580

Until it's trivial and well supported on macOS, I think we should allow `-DENABLE_OPENMP=OFF`.

papadop:
> > So I guess, we can replace this with a version check with find_package(OPENMP)
> 
> +1 for removing all other checks. Then we can say we just use c++17 plus (optionally) OpenMP
> 
> > Checking more thoroughly, it seems that VC++ is lagging behind !!!
> 
> Unfortunately we need Windows support, and we need it to work on Python. I'm not sure how good the support is for OpenMP + Windows nowadays, but I'm guessing "not good".
> 
> That being said, there is this infrastructure in conda-forge:
> 
> https://github.com/conda-forge/openmeeg-feedstock/blob/main/recipe/meta.yaml
> 
> The latest Windows build there did seem to have / use OpenMP:

I know we need windows. VC OpenMP support is at level 2.0, so probably no parallelization of range for and iterator loops, which are the features tested by the CHECK_CXX_OPENMP_SUPPORT macro. If we want to check for the need of this test, er need to see whether  OPENMP_RANGEFOR, OPENMP_ITERATOR and OPENMP_UNSIGNED are true on windows (which I do not believe as version 2.0 of OpenMP is found).

> https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=541794&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=66e4a587-a723-53e9-873c-6c9a2c5971ea&l=527
> 
> Looking at the latest wheel build, it also seems to have had it _EDIT: for windows_:
> 
> https://github.com/openmeeg/openmeeg/runs/7597415119?check_suite_focus=true#step:4:476

Not sure on what to look in these links.
 
> So macOS might actually be the issue here:
> 
> https://github.com/openmeeg/openmeeg/runs/7597415095?check_suite_focus=true#step:4:580
> 
> Until it's trivial and well supported on macOS, I think we should allow `-DENABLE_OPENMP=OFF`.

This is strange. I thought MacOS compiler was based ob clang/gcc, so should have OpenMP available... Maybe there is an extension to install. According to https://open-box.readthedocs.io/en/latest/installation/openmp_macos.html, we need to install libomp with Homebrew (or use the scikit-learns approach)...

larsoner:
> Not sure on what to look in these links.

Just that both of those windows build links show OpenMP being supported

>> Until it's trivial and well supported on macOS, I think we should allow -DENABLE_OPENMP=OFF.

> This is strange. I thought MacOS compiler was based ob clang/gcc, so should have OpenMP available... Maybe there is an extension to install.

So to me this does not sound trivial. I'd rather just leave the checks in so that on macOS people can compile without having to do these extra steps.

papadop:
Indeed, but on the other hand, we might add an OpenMP build in the ci builds....

larsoner:
You mean an OpenMP macOS build? I think it is using OpenMP on all other platforms. And it is using OpenMP on macOS on conda-forge, but that only builds with every release.

I can add a macOS+OpenMP (probably from `brew`) this week

papadop:
Done with PR #507. Modify it at will or merge it as you prefer.

larsoner:
@papadop I got rid of all checks other than OpenMP, feel free to review/merge

PR #409, opened 2020-01-04
New mesh i os

papadop:
This is a major refactoring of the mesh/geometry classes and in particular of their i/os. This homogeneized the multiple ways of importing meshes into a geometry (different part of the code were using different strategies). The code was difficult to understand and to maintain. Having multiple ways of doing things also made the python wrapping more difficult than necessary.

During this work, I also revamped some old and totally cryptic code for handling progressbar or some command line options.

The nice thing is that now direct python import of meshes works (I think). Quite heavily tested on fedora 31, let see what other arch will say (there is no much change in the C++ syntax, so I'm somewhat - and probably wrongly - optimistic).

Let's see what testers say....


.travis.yml
```diff
@@ -35,6 +35,7 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
           packages:
+            - g++-8
```
agramfort: @papadop can we avoid this?

.travis.yml
```diff
@@ -114,12 +115,13 @@ before_install:
...
   ############################################################################
-  # Install MKL (if necessary)
+  # Install MKL if necessary
   ############################################################################
 
   - |
     if [[ "${BACKEND}" == *"Intel"* ]]; then
-      curl -fsSkL https://raw.githubusercontent.com/openmeeg/ci-utils/master/travis/install_mkl.sh > install_mkl.sh && source ./install_mkl.sh
+      curl -fsSkL https://raw.githubusercontent.com/openmeeg/ci-utils/master/travis/install-mkl.sh > install_mkl.sh && source ./install_mkl.sh
+      echo "[[${MKLROOT} -- ${MKL_ROOT}]]"
```
agramfort: why do you need to change anything to MKL build system?

papadop: I discovered that they now provide apt repos.... This is probably faster and since it seemed there was problems with getting mkl, I tried this. It is simpler and should also be faster....

agramfort:
changing too many things at the same time (ie in one PR) is not ideal

can we do the code changes without having to update the CI system?
it took us weeks with @massich to make all this to work. Also having yet
another way to install things makes the number of options grow
exponentially.
Note that we need static builds for the release.

I am thinking that what you try here should be done upstream in
https://github.com/openmeeg/ci-utils/tree/master/travis
with some CIs setup there.

>

.travis.yml
papadop: Strange I do not remember having added this.... At least not on this branch..... I cloned from the master and it tells me that I'm in sync with it... And indeed this was added and then removed on the master.... So I guess we can avoid this.... But I do not understand where this change comes from....

agramfort:
this can explain maybe why you need to change anything to CIs?

papadop:
I needed to change just because there were spurious commits which were braking travis (from when I remade travis working).... This PR is cancelled. See PR #410 instead. I'm still trying the change about the mkl installation (it is a one character change since all is committed in the https://github.com/openmeeg/ci-utils/tree/master/travis 
and we can go back by just reverting the commit (since I added a new script instead of changing the old one).

PR #410, opened 2020-01-05
Mesh io cherrypicked

papadop:
Cancels PR #409 (there were spurious commits in that branch, so I cherrypicked the right commits).

This is a major refactoring of the mesh/geometry classes and in particular of their i/os. This homogeneized the multiple ways of importing meshes into a geometry (different part of the code were using different strategies). The code was difficult to understand and to maintain. Having multiple ways of doing things also made the python wrapping more difficult than necessary.

During this work, I also revamped some old and totally cryptic code for handling progressbar or some command line options.

The nice thing is that now direct python import of meshes works (I think). Quite heavily tested on fedora 31, let see what other arch will say (there is no much change in the C++ syntax, so I'm somewhat - and probably wrongly - optimistic).

Let's see what testers say, now that the travis is back (almost) to the one used in master....


papadop:
For the record:
- MacOS problems were due to brew delivering a python3 numpy where travis install a python 2.7 environment (brew link for python 3 was failing) -> fixed by using python3
- Ubuntu python problem -> seems to be related to a python 3.6 library problem -> attempt to use python 3.7.
- Windows problems -> simple problems in win32 specific code. Astonishingly this was the easier this time, but this is because python wrapping is not built. As we use python 3.6, I'd expect the same problems as with ubuntu. We have to test python wrapping with windows. Will do if/when I fix the ubuntu pb.

.travis.yml
```diff
@@ -114,12 +119,12 @@ before_install:
...
 
   ############################################################################
-  # Install MKL (if necessary)
+  # Install MKL if necessary
   ############################################################################
 
   - |
     if [[ "${BACKEND}" == *"Intel"* ]]; then
-      curl -fsSkL https://raw.githubusercontent.com/openmeeg/ci-utils/master/travis/install_mkl.sh > install_mkl.sh && source ./install_mkl.sh
+      curl -fsSkL https://raw.githubusercontent.com/openmeeg/ci-utils/master/travis/install-mkl.sh > install_mkl.sh && source ./install_mkl.sh
```
agramfort: is this really necessary? I just want to make sure it does not break the packaging of static binaries for mac and linux

papadop: I did not change anything in the cmake build, so it should change nothing in the packaging and especially on the static binaries. And certainly not for mac, for which I changed nothing....

agramfort:
Hopefully ok

But do we really have to do this in the PR?

papadop:
Well I'm trying to understand the failure. Indeed, it is bug research... It might be that finally all this is not strictly needed to the solution. On the other hand, if it's done, it's done... Not totally sure I want to undo all that stuff just to redo everything in another branch. Yes it would be cleaner, but No, I do not have the time...

But first, I need to understand th pb. I think swig is definitely a part of it (I found the commit that corrects the exact bug we encountered). Why cmake does not discovers numpy properly is my new puzzle....

papadop:
It is not strictly needed, but it is not either the bleeding edge (which is 3.16.2). We were just facing with a python3+numpy detection problem. I could certainly try to find a workaround, but my first choice is to see whether a newer version of the tools is not already fixing it....

papadop:
The real question in my view is that, with what I did, we dropped python 2 support. I think this is reasonable, but this is the real question.

Note that re-adding it might not be too difficult on the cmake side.... I'm more afraid with travis/appveyor, this is were I waste a lot of time. The interaction that forces me to push changes to see their effects is really not efficient and creates also all these spurious commits...

papadop:
Now that it is all green. Alex can you retest the python binding on your computer....
Another task (but I will do it in another PR, and probably not immediately) is to add a windows python build....

.travis.yml
```diff
@@ -26,6 +26,8 @@ matrix:
         - BACKEND=OpenBLAS
         - USE_VTK=ON
         - ENABLE_PYTHON=ON
+        - CMAKE_VERSION=3.16.2
```
agramfort: does it mean we don't have CI using older cmake now?

.travis.yml
```diff
@@ -26,6 +26,8 @@ matrix:
...
+        - SWIG_VERSION=4.0.1
```
agramfort: is this the minimal version we support now?

.travis.yml
agramfort: can we revert this now just to check it was not a necessary change?

.travis.yml
```diff
@@ -64,6 +65,7 @@ matrix:
 
     - name: openblas, vtk, full dep using brew
       os: osx
+      python: "3.7"
```
agramfort: did we bump dependency here? can you check that readme.rst file is still up to date?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.9)
+cmake_minimum_required (VERSION 3.12)
```
agramfort: I think if we do this we need to use cmake 3.12 in our CI

.travis.yml
papadop: I'm pretty sure it is not a necessary change (or at least, that the old way can be made to rework easily. I suspected changes in the paths on the intal web site). On the other hand, I'm sure it is a better way of doing things. In particular, mkl libraries are put where developpers will usually find it (ie not  in the openmeeg directory). Honestly, I do not see why you are so afraid.... The build has not been changed with this respect so if it was using static libraries, if they were not there it should have errorred... In additin, it is somewhat 

If you really want to test the old way, let's do a branch just to test that... and never integrate it. The new way is just better (at it is relies on the linux - here ubuntu - packaging system).

.travis.yml
papadop: For swig yes, definitely. swig 3 has a bug which was never corrected in the 3.x series.
For cmake as I said above, versions > 3.12 might be working.... I'm almost sure that in the debug process I used a 3.14 (before switching to 3.16) and it had the same symptoms 3.16 had before I found the real bug.... So going back to 3.14 is certainly doable.

.travis.yml
papadop: Indeed the rst is not up-to-date (Actually it was already out-of-date before these changes). A problem is that the instructions might differ with the fedora versions... Now, I must say, that README.rst does not really mention python for MacOSX.... It probably needs a small rewrite to better emphasize python support. I think on ubuntu, it works with python 3.6.3, so I do not know what you mean by bump the version.

Globally, probably, even python2 still works. It is only the cmake support for detecting python2 that is not there.... But really, do we want to support python2 ???

agramfort:
pushed changes to see how CIs react to changes of travis.

+1 to drop python 2 support. MNE does not support it anymore anyway.

>

agramfort:
thx @papadop !

I merged with rebase so you keep all the history.

PR #413, opened 2020-04-15
travis try

eolivi:


eolivi:
@agramfort @papadop @massich 
I've commited nothing in this PR and builds fail.
IMO because:
- brew update takes too much time. We could add `travis_wait` somewhere... or is this update really needed ?
- there are possibly memory leaks in our MathsIO , using a sanitizer give clues `CFLAGS=-fsanitze=address` (hdf5 is complaining "HDF5 error #000 in H5Gopen2()")
-

agramfort:
whatever you can do to help is great

Everytime I look into matio / hdf5 packaging issues I want to cry...

>

papadop:
I already had build problems with MacOSX. I have assumed that it was only because some travis builders are very slow (the time the build takes is not very consistent from one build to an another).
I will try a build with address-sanitizer locally later this week. I doubt that memory leaks can really make the build-and-test time explode.... but if there are memory leaks and we can correct them, that would be great.

agramfort:
gave up?

>

PR #416, opened 2020-04-18
start GH config

agramfort:


agramfort:
@papadop @eolivi I replicate with github actions https://github.com/openmeeg/openmeeg/pull/416/checks?check_run_id=598176331 the pb with libmatio hdf5 although I take packaged version from brew. A bug in matio that is not properly tested ?

agramfort:
as you can see @papadop I am going crazy because libmatio now uses a more recent of hdf5 on brew and matio is not properly tested (or we don't use it properly) so our mac os build is again broken...

agramfort:
@papadop if you can push here a build on ubuntu that uses the MKL package that would be very helpful. thanks

larsoner:
I am inclined to close this and start from scratch @agramfort

larsoner:
I am inclined to close this and start from scratch @agramfort

agramfort:
fair enough

PR #423, opened 2020-10-14
Factorize head matrix

papadop:
This PR factorizes the redundant code in HeadMat::HeadMat and HeadMatrix.
Note that this potentially also corrects some bug as the two versions were slightly different in the test for operatorD*.
 HeadMat::HeadMat was doing:  if ((mesh1!=mesh2) && (!mesh2.current_barrier()))
while
HeadMatrix was doing: if ((mesh1!=mesh2) && mesh2.current_barrier())

This is an untested cornercase (HeadMatrix is used only for CorticalMat and CorticalMat2), so I adopted the HeadMat::HeadMat variant, which I think is better tested. All tests are passing correctly.

### DIFF
--- a/OpenMEEG/include/mesh.h
+++ b/OpenMEEG/include/mesh.h
@@ -51,10 +51,12 @@ knowledge of the CeCILL-B license and that you accept its terms.
 #include <om_utils.h>
 
 #include <symmatrix.h>
+#include <block_matrix.h>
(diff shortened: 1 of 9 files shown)


papadop:
Grumpf.... I forked from the wrong branch....
The block matrix stuff is unrelated.

PR #427, opened 2021-07-06
Modified a typo in the `Vector::load` function

J-H-Cho:
Replaced `<<` with `>>` in the `Vector::load` function.


papadop:
Thank you for fixing this typo....

PR #433, opened 2021-12-14
Updates

papadop:
Several updates for new cmake, numpy, doxygen, ....
Improvement of the handling of exceptions in om_minverser.
No functional change otherwise.
All tests pass but CGAL build fails (but was failing before).

### DIFF
--- a/apps/minverser.cpp
+++ b/apps/minverser.cpp
@@ -59,7 +59,7 @@ getHelp(char** argv) {
 }
 
 int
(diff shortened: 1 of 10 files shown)


papadop:
Tested with cmake 3.22, doxygen 1.9.1 and numpy 1.21

papadop:
Removed some python tests which are no longer pertinent.

papadop:
Merged as I need it for sthg else.

PR #436, opened 2021-12-16
Accuracy improvement

papadop:
This is the improvement mentioned in #426 with all test file updated (all that were available from the tests). Note that only Head1.dsm is needed to have tests passing. The error that exceed the threshold are located (most surely, I have not checked exactly) close to the source dipole. Other files are regenerated because they should provide a more accurate baseline. Note that Head1.smooth is the only file not updated in the directory. This is because there is currently no tests that creates it (even though there are some remainders in TestHead.cmake).

We might investigate whether we could reduce the threshold with the increased accuracy. This is not done in this PR and not a priiority....


agramfort:
thx @papadop 

do you see an improved on MAG / RDM on the tests? if so maybe you can update the tolerance to highlight the accuracy gain?

papadop:
I have not tested that. I expect the accuracy to improve anecdotally in classical  MAG/RDM (I mean at surface points). As said in above comment, the value change is maximum close to the source (so inside the head). At best, with our tolerances which is 2e-5 of relative error IIRC, So, I do not expect a big change in MAG/RDM as scalp sensors were inside this tolerance.

I prefer we commit it as is and add an issue for further improvement/testing. But I agree that for a MAG/RDM computed on the whole head (which we never did before) that might make a difference.... For that I first need to compute the potential at all mesh nodes with an analytic code. This requires more time...

agramfort:
honestly you do the work so I won't object. But saying that I will add the
test later
is the best way to never find the time to do it. It's also a good way to
evaluate
the real impact to be able to communicate and document the benefit for
users.
Ideally we should have a changelog here... but again you do the work the
best
way you can. Merge if you want

papadop:
Better small steps than huge ones. This guaranties that things advance. Otherwise I fear that they still stay stationary for a long time. This simple patch was delayed for one year just because it was in a much bigger branch. There are many priorities.... But I will mark this in an issue.

PR #450, opened 2022-06-21
MAINT: Two more warnings

larsoner:
Takes care of:
```
[ 86%] Building CXX object wrapping/src/CMakeFiles/openmeeg.dir/CMakeFiles/openmeeg.dir/openmeegPYTHON_wrap.cxx.o
/Users/larsoner/python/openmeeg/build/wrapping/src/CMakeFiles/openmeeg.dir/openmeegPYTHON_wrap.cxx:6316:166: warning: data argument not used by format string [-Wformat-extra-args]
            std::snprintf(&buf[0], buf.size(), "Wrong dtype for triangles array (only 32 or 64 int or uint supported), got type '%c%d'", descr->kind, descr->elsize, type_num);
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                              ^
/Users/larsoner/python/openmeeg/build/wrapping/src/CMakeFiles/openmeeg.dir/openmeegPYTHON_wrap.cxx:45802:7: warning: delete called on 'OpenMEEG::GeometryIO' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]
      delete arg1;
      ^
```
Also renames the GitHub action to make more sense, and turns on warnings-as-errors on macOS. Just Linux after that (which I'll try to take care of when on my Linux machine sometime).

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -67,11 +67,11 @@ jobs:
         else
           BUILD_DOCS=OFF
         fi
-        # Use -Werror on Windows, eventually should use everywhere
-        if [[ ${{matrix.os}} == 'windows*' ]]; then
-          ENABLE_WERROR=ON
-        else
+        # Use -Werror on Windows and macOS, eventually should use on Linux
+        if [[ ${{matrix.os}} != 'ubuntu*' ]]; then
           ENABLE_WERROR=OFF
+        else
+          ENABLE_WERROR=ON
         fi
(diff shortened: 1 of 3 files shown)


larsoner:
🦄 Green on the first try! I don't care if it's only +7/-5, that's a clear win in my book

PR #453, opened 2022-06-22
MAINT: Build wheels

larsoner:
This assumes that cmake has been used to build the lib. After that, `setup.py` can be used to build a wheel. Then hopefully it can be used to delocate the wheel....


larsoner:
All good, wheels build and at least import on all systems after delocate (macOS) / delvewheel (Windows). Can't test auditwheel (Linux) here because the CIs are too new, but I'd expect CIBUILDWHEEL's docker system to take care of this easily, and Linux to be the least problematic platform anyway.

One remaining issue is that the wheels are produced with the `.any` name, which is not correct. But I'll add that to the TODO list...

PR #461, opened 2022-07-09
BUG: Fix bug with dir perms

larsoner:
When I nested `openmeeg` I created a bug where the Python files couldn't actually be accessed. This should fix it, and adds a test that the `install` step produces an importable Python binary and runnable executables.


larsoner:
Okay this fixes the problem I introduced and checks that `cmake --install` works on CIs (up to issues with path problems / DLL resolution on Windows that I put a hack workaround in for) so I'll go ahead and merge!

PR #465, opened 2022-07-12
MAINT: Allow use of SWIG separately

larsoner:
This PR makes it so that SWIG can be used separately via `setup.py` rather than via cmake. This required:

- [x] Exporting to `exports/` so that the `_Export.h` files could be more easily / in a more standard way included in the `install` step
- [x] Putting some defines in the make cmake namespace (if that's what it's called) so that they could be recycled in `OpenMEEGConfigure.h.in` / `configure_file` via `#cmakedefine`

Now locally I was able to:
```
PYTHON_OPT=-DENABLE_PYTHON=OFF ./build_tools/cmake_configure.sh
cmake --build build
cmake --install build --prefix install
export OPENMEEG_INCLUDE=$PWD/install/include/OpenMEEG
export OPENMEEG_LIB=$PWD/install/lib
cd wrapping/python
OPENMEEG_USE_SWIG=true python setup.py develop --user
```
and things worked. So now my plan is:

- [x] Push existing changes, get CIs green
- [x] Ensure that the SWIG-built wheels do not have `any` in the name
- [x] Actually use SWIG and `-DENABLE_PYTHON=OFF` in place of `-DENABLE_PYTHON=ON` in some jobs
- [x] Reenable commented-out jobs

This should put us on the verge of being able to build wheels.


.github/workflows/build_and_test.yml
```diff
@@ -344,13 +379,10 @@ jobs:
...
         ./install/bin/om_minverser --help
-        if [[ "${{matrix.python}}" == 'python'* ]]; then
+        if [[ "${{steps.env-vars.outputs.python_type}}" == 'cmake' ]]; then
           PYVER=$(python -c "import sys; print('.'.join(sys.version.split('.')[:2]))")
           if [[ "${{matrix.os}}" == 'windows'* ]]; then
-            export PYTHONPATH=$PWD/install/lib/site-packages
-            curl -LO https://github.com/lucasg/Dependencies/releases/download/v1.11.1/Dependencies_x64_Release.zip
-            unzip Dependencies_x64_Release.zip -d Dependencies
-            # TODO: This should not be necessary but let's see if it makes things green
+            # Ideally this would not be necessary but it makes things work
```
larsoner, line 385:
```suggestion
            export PYTHONPATH=$PWD/install/lib/site-packages
            curl -LO https://github.com/lucasg/Dependencies/releases/download/v1.11.1/Dependencies_x64_Release.zip
            unzip Dependencies_x64_Release.zip -d Dependencies
            # Ideally this would not be necessary but it makes things work
```

.github/workflows/build_and_test.yml
```diff
@@ -360,11 +395,38 @@ jobs:
...
+          export OPENMEEG_INCLUDE=$PWD/install/include/OpenMEEG
+          export OPENMEEG_LIB=$PWD/install/lib
+          export OPENMEEG_USE_SWIG=1
+          if [[ "${{matrix.os}}" == 'windows'* ]]; then
+            export PATH="$PWD/build:$PWD/install/bin:$(cygpath -u $CONDA_PREFIX)/Library:$(cygpath -u $CONDA_PREFIX)/Library/lib:$PATH"
+            export OPENMEEG_LIB=$(cygpath -w $OPENMEEG_LIB)
+            export OPENMEEG_INCLUDE=$(cygpath -w $OPENMEEG_INCLUDE)
+            export CPATH="$OPENBLAS_INCLUDE:$CPATH"
+            echo "PATH=$PATH"
+            echo "CPATH=$CPATH"
```
larsoner, line 412:
```suggestion
            echo "CL=$CL"
```

.github/workflows/build_and_test.yml
```diff
@@ -360,11 +395,38 @@ jobs:
...
-        cd build/wrapping/python
+        if [[ "${{steps.env-vars.outputs.python_type}}" == 'setuptools' ]]; then
+          export OPENMEEG_INCLUDE=$PWD/install/include/OpenMEEG
+          export OPENMEEG_LIB=$PWD/install/lib
+          export OPENMEEG_USE_SWIG=1
+          if [[ "${{matrix.os}}" == 'windows'* ]]; then
+            export PATH="$PWD/build:$PWD/install/bin:$(cygpath -u $CONDA_PREFIX)/Library:$(cygpath -u $CONDA_PREFIX)/Library/lib:$PATH"
+            export OPENMEEG_LIB=$(cygpath -w $OPENMEEG_LIB)
+            export OPENMEEG_INCLUDE=$(cygpath -w $OPENMEEG_INCLUDE)
+            export CPATH="$OPENBLAS_INCLUDE:$CPATH"
```
larsoner, line 410:
```suggestion
            if [[ "${{matrix.windows_compiler}}" == 'msvc'* ]]; then
              export CL="/I$OPENBLAS_INCLUDE /std:c++17 /VERBOSE"
            fi
```

larsoner:
Okay this works except for persistent windows SDK _FComplex issues which hopefully won't be present with cibuildwheel so I'll go ahead and merge!

PR #470, opened 2022-07-18
MAINT: Reenable codecov

larsoner:
Should restore coverage


larsoner:
`codecov/project/library Successful in 1s — 78.98% (target 90.00%)` In it goes!

PR #477, opened 2022-07-24
MAINT: Reenable pypy wheels and use oldest-supported-numpy

larsoner:


.github/workflows/build_and_test.yml
```diff
@@ -208,7 +208,7 @@ jobs:
           echo "Installing oldest supported NumPy"
           NUMPY_PIP=oldest-supported-numpy
```
larsoner, line 209:
```suggestion
          IS_PYPY=$(python -c "import sys; print(int('pypy' in sys.implementation.name))")
          if [[ "$IS_PYPY" == "1" ]]; then
            NUMPY_PIP="numpy==1.23.0"
          else
            NUMPY_PIP="oldest-supported-numpy"
          fi
```

wrapping/python/pyproject.toml
```diff
@@ -4,11 +4,13 @@
...
     "wheel==0.37.0",
-    "numpy"
+    "oldest-supported-numpy; platform_python_implementation!='PyPy'",
+    "numpy==1.23.0; platform_python_implementation=='PyPy'",
 ]
 
 [tool.cibuildwheel]
-skip = "cp36-* *-manylinux_i686 *_ppc64le *_s390x *-musllinux* *-win32 pp*"
+# TODO: only 3.8 until NumPy supplies wheels for 3.9
+skip = "cp36-* *-manylinux_i686 *_ppc64le *_s390x *-musllinux* *-win32 pp37* pp39*"
```
larsoner, line 13:
```suggestion
skip = "cp36-* *-manylinux_i686 *_ppc64le *_s390x *-musllinux* *-win32 pp37* pp39* pp*-win_*"
```

PR #488, opened 2022-07-27
Please fix the bug in Geomfile.h

vhs304305:
Please fix the bug in Geomfile.h, line 213, 214

from
bool has_meshfile;
bool has_meshsection;
to
bool has_meshfile = false;
bool has_meshsection = false;

They were used later but not initialized, making reading version 1.0 wrong (incompatible).


agramfort:
@vhs304305 you did not branch from latest master branch. Your diff is too big

eolivi:
close as done in https://github.com/openmeeg/openmeeg/pull/491

Issue #487, opened 2022-07-26
Bug in Geomfile.h


vhs304305:
Please fix the bug in Geomfile.h, line 213, 214

from
        bool has_meshfile;
        bool has_meshsection;
to
        bool has_meshfile = false;
        bool has_meshsection = false;

They were used later but not initialized, making reading version 1.0 wrong (incompatible).

larsoner:
Do you want to make a PR to fix it?

papadop:
Thank's for the report and for the fix.

For the record. The variables are normally set only inside a block conditioned by  the file version (version_id==VERSION11), (in io_utils::match_optional) but indeed used after that block so can end up being used without being initialized.

I wonder how it passed tests. This certainly means that we no longer have tests involving the old format, which is bad because it means we can regress without knowing it.

So in addition to fix the initialization, we also need to add a test for the old file format if we do not have one, and if we have one understand why we did not caught this error.

papadop:
And by the way, thank you @larsoner for picking up the task of making the checks work again. This is very much appreciated... I tried to do it in December, and I really know how much work is needed to make all the arch/OS/options to work. So I take this occasion to thank you for this great work.

larsoner:
It was a... "fun" journey @papadop :) If you want to see the plan I've been following it's in https://github.com/openmeeg/openmeeg/issues/449

In particular you are probably interested in and could help with the commits of yours I had to revert to get things green:

https://github.com/openmeeg/openmeeg/pull/457#issuecomment-1171916883

papadop:
Yeah, I discovered it (read it a little bit quickly) after sending the post. I will try to  re-add the reverted commits one by one.

papadop:
This should be fixed with the merge of PR #491 and a test case to avoid any future regression is introduced with PR #494 which is ready to be merged. Thank you again for reporting the issue. I'm closing it as I believe that it is now fixed.

As a final note, please note that in some undefined future, we might stop supporting the old file format, so it is best if you could convert to the new format 1.1 (as said in the warnings you should get when using this old file format with anay openmmeg command). This is more future proof....

PR #491, opened 2022-07-29
Solve undefined behavior caused by uninitialized variable when readin…

eolivi:
for solving https://github.com/openmeeg/openmeeg/issues/487


larsoner:
As @papadop mentioned in another issue, it would be good to have a test that would have caught this. But given it's a corner case and it might be tough, we can do it later if someone finds time.

Thanks @eolivi !

PR #492, opened 2022-07-29
Slightly improve the previous fix.

papadop:
Just completely avoid to set up the random noise generator if it is not to be used.


larsoner:
Thanks @papadop !

PR #493, opened 2022-07-29
Add missing exception for block API.

papadop:
Un-revert the Block API (commit 1). This one should not be problematic as it just adds new classes.

### DIFF
--- a/OpenMEEGMaths/include/Exceptions.H
+++ b/OpenMEEGMaths/include/Exceptions.H
@@ -12,11 +12,14 @@
 namespace OpenMEEG {
     namespace maths {
 
+        class Range;
+        std::ostream& operator<<(std::ostream& os,const Range& r);
+
         typedef enum { UNEXPECTED = 128, IO_EXCPT,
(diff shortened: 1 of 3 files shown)


larsoner:
MSVC/windows failures are real

https://github.com/openmeeg/openmeeg/runs/7579069788?check_suite_focus=true

larsoner:
@papadop can you please revert? I think it's critical we keep CIs green, and they were red here

agramfort:
@papadop did you just merge a red PR consciously ?

papadop:
I do not understand. All tests were green after the third commit....  Are you sure you updated the github page ?
Windows build were indeed failing before the last commit but all passed after it...

I have some troubles because I do no longer receive the notifications on the issues or the cis. I have not changed anything and the github configuration says that I should receive them. Any idea ? I still receive the merge and commit notifications.
This is why I did not see the last messages posted here and send the emails....

papadop:
And the issue was actually closed. Resent as PR #498. Let's see what testers will say.

papadop:
And no I will not 'merge a red PR consciously ?'.
I appreciate enough the fact that they work again (and thanks again).

larsoner:
@papadop it shows up as red above, even now:

Screenshot from 2022-07-29 16-07-10


But I think this is just a GitHub issue since looking at the actual statuses, everything is green (or neutral/not run). So I think we're actually okay here!

PR #496, opened 2022-07-29
Add block matrix APIs.

papadop:
Add back the block matrix API. Let's see if the checkers pass. Normally the block API is not used yet, so no new failures should be seen, but there are new tests...


larsoner:
For example regarding the "standard" flow @papadop I am tempted to delete this branch because it's stale *and* on `openmeeg`, but I don't know if you're done with it so I can't delete it. If it's on your fork, you get to decide how to keep it tidy (or not) rather than having to "share the space" with everyone else.

PR #494, opened 2022-07-29
Add legacy file format test

papadop:
Add a test and improve slightly the message given when an unknown file format is provided (identified by the file extension).

PR #498, opened 2022-07-29
Block api exceptions

papadop:
Retry

### DIFF
--- a/OpenMEEGMaths/include/Exceptions.H
+++ b/OpenMEEGMaths/include/Exceptions.H
@@ -12,11 +12,14 @@
 namespace OpenMEEG {
     namespace maths {
 
+        class Range;
+        std::ostream& operator<<(std::ostream& os,const Range& r);
+
         typedef enum { UNEXPECTED = 128, IO_EXCPT,
                        BAD_FILE, BAD_FILE_OPEN, BAD_CONTENT, NO_SUFFIX, BAD_HDR, BAD_DATA, BAD_VECT, UNKN_DIM, BAD_SYMM_MAT,
-                       BAD_STORAGE_TYPE, NO_IO, MATIO_ERROR, UNKN_FILE_FMT, UNKN_FILE_SUFFIX, NO_FILE_FMT, UNKN_NAMED_FILE_FMT,
-                       IMPOSSIBLE_IDENTIFICATION } ExceptionCode;
-
+                       BAD_STORAGE_TYPE, NO_IO, OVERLAPPING_RANGES, NON_EXISTING_BLOCK, MATIO_ERROR, UNKN_FILE_FMT,
+                       UNKN_FILE_SUFFIX, NO_FILE_FMT, UNKN_NAMED_FILE_FMT, IMPOSSIBLE_IDENTIFICATION
+                     } ExceptionCode;
 
         class Exception: public std::exception {
         public:
@@ -205,6 +208,36 @@ namespace OpenMEEG {
 
             static std::string message(const std::string& file,const Mode& mode) {
                 return std::string("Unable to find ")+((mode==READ) ? "reader" : "writer")+" for file "+file+".";
+            }
+        };
+
+        struct OPENMEEGMATHS_EXPORT OverlappingRanges: public Exception {
+
+            OverlappingRanges(const Range& r1,const Range& r2): Exception(message(r1,r2)) { }
+
+            ExceptionCode code() const throw() { return OVERLAPPING_RANGES; }
+
+        private:
+
+            static std::string message(const Range& r1,const Range& r2) {
+                std::ostringstream ost;
+                ost << "Ranges " << r1 << " and " << r2 << " are overlapping blocks in a block matrix.";
+                return ost.str();
+            }
+        };
+
+        struct OPENMEEGMATHS_EXPORT NonExistingBlock: public Exception {
+
+            NonExistingBlock(const unsigned ind): Exception(message(ind)) { }
+
+            ExceptionCode code() const throw() { return NON_EXISTING_BLOCK; }
+
+        private:
+
+            static std::string message(const unsigned ind) {
+                std::ostringstream ost;
+                ost << "Attempt to access a non-existing block at index " << ind << '.';
+                return ost.str();
             }
         };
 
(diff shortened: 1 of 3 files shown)


agramfort:
@papadop can you open PRs from your fork as suggested by @larsoner earlier today ? 🙏

papadop:
I did not see the @larsoner suggestion. Where is it ?
Honestly, it is a burden to me to do that. It requires synchronizing multiple repositories and I have too often struggle to maintain all the repos/branches in sync. I may not master git well enough, but it really makes me loosing time. And I do not see the real drawback of using the main repository.

papadop:
Now, I do not understand. As far as I can tell, all checks are OK right ?
Then you told me that commits have crosses, but I never linked this crosses with failures. Is this a new gitflow thing ?
To me the PR is OK to be merged.... Because, this is exactly the situation in which I merged earlier this afternoon....
So is this branch mergeable or not ? And if not where can I see what is the problem ?

larsoner:
> I did not see the @larsoner suggestion. Where is it ?

I snuck it in after a merge

https://github.com/openmeeg/openmeeg/pull/494#issuecomment-1199439784

> Honestly, it is a burden to me to do that. It requires synchronizing multiple repositories and I had too often struggle to maintain all the repos/branches in sync. I may not master git well enough, but it really makes me loosing time. And I do not see the real drawback of using the main repository.

There are at least a few. The biggest to me is that it deviates from the more standard way of doing things nowadays, which means that everyone familiar with this way of doing things will be confused by it and have to adapt to it. I see this as a drawback for you personally as well @papadop, as it means you miss out on learning something potentially new and useful that you could bring to other projects you could contribute to! (Or to put it differently, if you learn this de facto standard here, it applies elsewhere almost immediately, so it's a useful thing to know.)

For other contributors, it also means that anytime they do a `git fetch upstream`, they fetch not just latest `main` but also all of your branches. This does not scale well as more and more contributors are added. It's not a huge concern since that's probably less than 5 currently, but still worth thinking about.

> It requires synchronizing multiple repositories 

FWIW you only really need to keep track of two forks/repos: yours (which you'd create called `origin`), and a new one called `upstream` (which would be this one). And you'd only need to interact with the `upstream` ever to update to latest `main`. So in other words, starting from what you have now with this repo as `origin`, since [you already forked](https://github.com/papadop/openmeeg) you could:

1. `git remote rename origin upstream` -> set this repo to be named `upstream`
2. `git remote add origin git@github.com:/papadop/openmeeg.git` -> add your fork named as `origin`

Then you're configured properly. Assuming you haven't made a `main` branch yet (I renamed `master` at some point) you would do:
```
git fetch upstream
git checkout -b main upstream/main
```
Then anytime you wanted your `main` branch to look like upstream/main, you'd update it:
```
git checkout main
git pull --ff-only upstream/main
git push origin main  # if you want but you don't really need to
```
and to create PRs you'd do something a lot like what you're already doing:
```
git checkout main
git checkout -b new-feature
<hack>
git commit -am 'ENH: I added a thing'
git push origin -u new-feature
<go on GH and open PR>
```
In other words, you more or less just need to add `upstream` and keep your local (on your HD) `main` branch in sync with it every once in a while. Other than that you interact with `origin` as usual.

> So is this branch mergeable or not ? And if not where can I see what is the problem ?

I think it probably is -- GitHub is doing something weird where these are green, but the box next to your commit above itself is red (which is all you see after a merge):

Screenshot from 2022-07-29 16-20-56

I'll see if this is some problem with branch protection settings or something, then I'll merge!

larsoner:
Thanks @papadop . I deleted the branch since it has been merged

papadop:
Honestly, it is not as if I had not tried (and do it on repositories on which I have no particular rights). If I'm telling you it generates more work, believe me it's from experience. I may not master git well enough, but given the fact that I'm using it in bursts, it is never becoming a routine to me (all the more that there are a lot of small details and complexities). But its not just me: on occasion, I asked to seasoned software engineers to show me how to solve the situation I was in, and it never succeeded better... So maybe I'm doing something wrong, but it is clearly not as simple as you say in my situation.

My main problem is not covered by the work flow you describe. After a while, I'm coming back a a branch forked from my own master (or main), and both the branch and the master need to be updated from the upstream. This is usually when I encounter problems (eventually having to solve conflicts numerous times). 

I can only do OpenMEEG (and other) programming in short bursts of time and I have branches that may last for quite a long time before I can finish them (usually they remain local). The number of times I had to re-create a branch by cherry-picking each commit and re-applying it and correcting all conflicts after going back to a branch after a few months is a reality that makes me "loose" half a day or more of work each time and is a real burden to me.

I will try (after the currently existing branches) one more time, but honestly if I have to spend even one hour to solve git subtleties, I will probably go back to a simpler workflow....

By the way, can you explain the rationale of changing master to main ? I'm not opposed to it (though it will introduce some more git commands for me to refresh all my local branches and copies of the repository), I just want to understand, because I see so many variations (some keep master form releases and name the "master" the development branch (a la gitflow), some have other conventions).

Finally, is there a place where I can understand all the checks that are done (beside arch/os), e.g. what is the difference between a Test/buidl and a cibuildwheel or the various tag names (setuptools, numpyold, ....). I can probably get the information by digging in your recent changes, but I will be grateful if there was a synthetic description somewhere to catch up.

Thank's for merging, let's now try the next revert....

agramfort:
@papadop you don't need to change much to your workflow.
Basically your master branch should never be different from the
master branch upstream.
It's just when you push a branch that is not master (or main if we rename)
that you need to push
to your fork.

git push <name of your fork> <branch name>

regarding renaming master to main it's now the new default in git. Main is
a more politically correct name than master...

larsoner:
> After a while, I'm coming back a a branch forked from my own master (or main), and both the branch and the master need to be updated from the upstream. This is usually when I encounter problems (eventually having to solve conflicts numerous times).

I think this will be the same in your simpler workflow and the more standard GitHub fork-based workflow I described. Branches get out of date if other branches get merged in the meantime; this can cause conflicts; and when it does, they need to be updated by `merge`, `rebase`, or starting over and `cherry-pick`ing -- and it's a pain in any of those cases unfortunately. In this sense I don't think the standard fork-based workflow makes this problem harder.

> is there a place where I can understand all the checks that are done

Can you look at https://github.com/openmeeg/openmeeg/pull/500 and mark it as "merge when green" if you're happy, and comment directly on the PR if things are still not clear? Happy to add more text if it would help

Issue #495, opened 2022-07-29
MAINT: Replace release url with GH releases


larsoner:
@papadop your comment in https://github.com/openmeeg/openmeeg/issues/331#issuecomment-1199444834 reminded me -- if you feel like doing some CI infrastructure work, we should replace https://files.inria.fr/OpenMEEG/download/ with adding assets directly to the GitHub releases, e.g. how it's done here:

https://github.com/mne-tools/mne-installers/releases/tag/v1.0.3
https://github.com/mne-tools/mne-installers/blob/59ca501957c2ff54f359cacd2375c9b87cbd58ed/.github/workflows/build.yml#L436-L500

Basically we "just" need to have a CI run that builds the binaries for each platform (we are probably close on the 18.04 run for linux, or the cibuildwheel runs for even better backward compat), outputs them as an artifact, then use the action to attach them to the release

- [ ] Replace https://files.inria.fr/OpenMEEG/download with GH releases that are build on commit/release
- [ ] Update `openmeeg.github.io`

larsoner:
We might want to do a build separate from `cibuildwheel` that uses all static linking. The `cibuildwheel` uses dynamic and then vendors the libraries.

Alternatively, we can static linking working properly in `cibuildwheel`, but this was a pain at some point when I tried it.

Either way I think we get close to what we want with `-DENABLE_PACKAGING=ON` and `--target package` in our `cmake` command.

PR #500, opened 2022-07-30
DOC: Document CIs

larsoner:
Overdue CI overview

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -1,3 +1,15 @@
+# This CI run builds wheels using the cibuildwheel infrastructure.
+# This means it:
+# - Uses as old dependencies/infrastructure as possible:
+#   - An old manylinux docker container on Linux
+#   - An old MACOSX_DEPLOYMENT_TARGET on macOS
+#   - An old SYSTEM_VERSION target on Windows
+# - Builds OpenMEEG
+# - Creates wheels
+# - Delocates them
+# - Uploads them to TestPyPI (main) or PyPi (releases)
+# Most of the config lives in `pyproject.toml` and build_tools/cibw_* scripts.
+
 name: cibuildwheel
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,3 +1,6 @@
+# This CI run does core building+testing of OpenMEEG using various OSes
+# and options.
+
 name: Test
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
@@ -27,6 +30,18 @@ jobs:
         # (https://github.com/microsoft/vcpkg/issues/25176), but it's unclear
         # if this will really help anything. SciPy's should be well tested
         # already.
+        #
+        # By default, all builds use:
+        # - OpenBLAS
+        # - dynamic BLAS linking
+        # - cmake to build Python bindings
+        # - Python 3.10
+        # - The oldest supported version of numpy (numpyold)
+        # - MSVC on Windows (if relevant)
+        #
+        # Deviations from this are noted by additions to the `include`
+        # params. Defaults are set up properly in the `env-vars` step,
+        # whose `output`s should be used by subsequent steps.
         include:
         - os: ubuntu-20.04
           blas: OpenBLAS


agramfort (approved):

papadop:
Thank you. Just some clarifications:

> # Currently, VTK is only installed+tested on Ubuntu
        # Someday we could use OpenBLAS for Windows once vcpkg sorts things out
        # (https://github.com/microsoft/vcpkg/issues/25176), but it's unclear
        # if this will really help anything. SciPy's should be well tested
        # already.

I do not see any VTK related stuff in build_and_test.yml.
Similarly, I do see Windows build with OpenBLAS. Is the comment still true ?
I do not understand the SciPy comment either. Do you mean Numpy ?

papadop:
Is there any MKL test yet ?

papadop:
I still do not quite understand what are the "wheels" tests ? Are these specific to python ? Are these for old arch/OSes ? In what do they differ from the "build and test ". Is this just a different testing infrastructure ? Sorry, this is probably trivial, but I have not followed this....

.github/workflows/cibuildwheel.yml
```diff
@@ -1,3 +1,15 @@
+# This CI run builds wheels using the cibuildwheel infrastructure.
+# This means it:
+# - Uses as old dependencies/infrastructure as possible:
```
larsoner, line 3: > Are these for old arch/OSes ?

This comment was meant for this part. Yes it's for as far backward in OS compat as we can get

.github/workflows/cibuildwheel.yml
```diff
@@ -1,3 +1,15 @@
...
+#   - An old manylinux docker container on Linux
+#   - An old MACOSX_DEPLOYMENT_TARGET on macOS
+#   - An old SYSTEM_VERSION target on Windows
+# - Builds OpenMEEG
+# - Creates wheels
+# - Delocates them
+# - Uploads them to TestPyPI (main) or PyPi (releases)
```
larsoner, line 10: > Are these specific to python ?

This comment is meant to answer this part. More or less, yes.

*But*, I think we could use it to take care of https://github.com/openmeeg/openmeeg/issues/495 as well. I think if we `-DENABLE_APPS=ON` in the wheel build instead of `-DENABLE_APPS=OFF`, we'll have all the files we need and it's "just" a packaging problem at that point.

larsoner:
> Is there any MKL test yet ?

Yes on each OS, see this comment, the `blas:` arg, and during CI runs the "MKL" that shows up in CI names:

https://github.com/openmeeg/openmeeg/blob/18f062df4d64b36dd28824c56087b7d56f82a30b/.github/workflows/build_and_test.yml#L76

> Similarly, I do see Windows build with OpenBLAS. Is the comment still true ?

Yes we use Windows+OpenBLAS. The comment is just to say that someday we could use `vcpkg` to do it instead if we want (but it's broken there).

> I do not understand the SciPy comment either. Do you mean Numpy ?

I mean this, which I think was a joint effort between SciPy + NumPy:

https://github.com/MacPython/openblas-libs

> I do not see any VTK related stuff in build_and_test.yml.

This I think I errantly removed, I'll make a commit to add it back in

papadop:
Thank's for the answers.
Can we clarify the comment as: 

> # Someday we could use OpenBLAS for Windows _with vcpkg_ once vcpkg sorts things out

papadop:
On MKL, would it be possible to have the tag MKL instead of mkl-findblas ? This is just for homogeneity with the OpenBLAS tag....

larsoner:
> Similarly, can you clarify the Scipy comment ? I'm not sure I understand the openblas-libs stuff. Is this to base a build on the latest OpenBLAS ?

Can you look at https://github.com/openmeeg/openmeeg/pull/505 ? I've actually already improved these comments there

> On MKL, would it be possible to have the tag MKL instead of mkl-findblas ? This is just for homogeneity with the OpenBLAS tag....

So the history here is that OpenMEEG had a bunch of machinery/code for figuring out how to get MKL to work properly. `mkl-findblas` ignores all of this and lets CMake's internal FindBLAS routines do all the work, rather than our internal code. You can see this for example here:

https://github.com/openmeeg/openmeeg/blob/main/cmake/FindBLASImplementation.cmake#L5

I'd be +1 for removing all of our custom stuff and just making `mkl` mean `mkl-findblas` (i.e., let `cmake` do all the work), but I was hesitant to remove all of this code since it seems like someone worked hard on it for a reason. But it's entirely possible that with modern `cmake` their support has improved and none of this code is needed anymore...

papadop:
+1 for removing the code too. I think that cmake support was just not yet up to what we needed. It seems that besides your recent change, I'm the one that worked on this lately (but before I'm pretty sure that @eolivi and @massich worked on it too).
MKL has always been complicated to handle and maintain. If cmake support is now up to the task, I'm perfectly fine with removing the specific code from OpenMEEG. The only check I would do is to verify that findblas-mkl selects the mkl version 
(e.g. parallel vs sequential) that is the most efficient, i.e. that the build with mkl-findblas is as fast as the one selected with the original code on the testsuite.

agramfort:
if everything works without then no pb from my end to remove some code

PR #501, opened 2022-07-30
New block api

papadop:
This PR adds the block matrix classes (full and symmetric). Those are not yet used except in a new test  which just verifies that everything compiles and runs without errors (the output of the test is not yet checked for correctedness).


papadop:
This passes all tests. Merging.

PR #502, opened 2022-07-30
Introduce range methods.

papadop:
This starts to introduce some range methods into the mesh class. They are not yet used, so it should not perturb build and testing.


papadop:
All tests have passed. Merging.

PR #503, opened 2022-07-30
OpenMEEG block api

papadop:
This adds the block API to OpenMEEG. This also reorganizes some pieces of code notably by replacing trivial classes with functions and the ways operators are defined.


papadop:
All tests passed. I think all reverted commits are now in.

larsoner:
Thanks @papadop !

FYI the checkmarks are now also showing up as green. I think it's because on this PR LGTM gave green for both languages, on the ones that were red the Python one showed up as Neutral. It seems like a GitHub issue that a Neutral check causes a red icon (and there is no way to tell LGTM to treat neutral as good when reporting, or GitHub to treat neutral as good for the commit status icon) but we can live with it by deaubl-checking the PR CIs

PR #377, opened 2019-06-14
WIP: simplify clean-test

massich:

PR #504, opened 2022-07-31
Massich clean test

papadop:
This PR takes over PR #377 using the same ideas and going a little bit further but moving all the test logic into the tests directory (I do not like much the idea of moving the cmake files into the cmake directory. This is the only departure from the @massich ideas, otherwise following what I think he was proposing).


papadop:
I forgot to say that some tests are now better ordered. The load_geometry like tests are now done before the OpenMEEG tests (even if this is not enforced by dependencies). Another small logic change which is invisible is that those load_geom type of tests (and the associated targets) were done systematically before, but are now conditioned by BUILD_TESTING. But as BUILD_TESTING is always true, this is strictly equivalent for now.

larsoner:
Is it worth explicitly testing this by adding another step in `build_and_test.yml` than does something like
```
# Ensure clean-test works
test -f <some_file>
<clean test>
test ! -f <some_file>
```
?

papadop:
Do you mean just for testing clean-test ? Not sure what it would bring.
I leave the decision to you...

larsoner:
> Do you mean just for testing clean-test ? Not sure what it would bring.

The idea is the same as any test-driven development -- add some test that ensures what you wrote does what it claims to do, and ensure that we don't break it in the future.

papadop:
But if we start to test that tests are doing what they are supposed to do, when do we stop ? Honestly, unless @agramfort votes otherwise, I do not think this is useful. I'll leave it to someone to add this eventually.

larsoner:
> But if we start to test that tests 

My thinking was that this was a "clean" step, so we'd be checking that the clean step actually cleaned. But if you're not convinced I can live without it

larsoner:
Thanks @papadop !

PR #506, opened 2022-07-31
Minor reorganization of the code and comment update.

papadop:
Just a small update of a comment and an inlining of a private method used only once/


papadop:
All tests passed. Merging.

PR #514, opened 2022-08-01
WIP: Try forcing static on Windows

### LINKED ISSUE #510: BUG: Fix static linking of OpenBLAS on Windows
It would good to allow the windows build to actually link statically to OpenBLAS -- it seems to still pick up the `.dll` even when the static build is requested.

One maybe easy thing to try is removing the `.dll`s from the static build to force cmake+MSVC to do the right thing here (?).

larsoner:
Tries the brute force approach idea in https://github.com/openmeeg/openmeeg/issues/510

CI skips should be uncommented before merge

Closes #510


larsoner:
Okay it was not so trivial. This will probably require actually debugging in a Windows boot. I'll close for now

PR #516, opened 2022-08-02
Correct the syntax of env varaiables which must be separated by a ;

papadop:
Environment variables must be separated by a semi-column. See https://stackoverflow.com/questions/48954920/trouble-setting-environment-variables-for-ctest-tests
I chose to break the statement in to, because this avoid a too long line, and I find it clearer.
This is more friendly for developers instead of asking them to define the variable.

I found this while testing python tests on fedora.
I think that this may allow the some simplifications in build_and_test.yml and potentially in cibw_test_command.sh (but I do not understand this one well enough, I leave this for @larsoner, if it is relevant.


larsoner:
> this may allow the some simplifications in build_and_test.yml and potentially in cibw_test_command.sh

It should allow for not setting `OPENMEEG_DATA_PATH` when using `ctest`-based tests, but it will still be needed for any direct `pytest` calls (which is what `cibw_test_command.sh` uses. So I think you've removed it in the correct place already

larsoner:
Thanks @papadop for cleaning up my mistakes here and elsewhere !

papadop:
No problem. We all make mistakes. I have found a bunch of them with your WERROR option in some code I committed, so as we say in french:
"C'est un prêté pour un rendu...." (basically it's an exchange).

PR #286, opened 2018-03-27
[No-MRG] Add cgal with PPA

massich:

PR #285, opened 2018-03-27
[WIP] Add cgal with conda

massich:
[Add CGAL to OpenMEEG](https://trello.com/c/UdZkBmtY/90-add-cgal-to-openmeeg)

PR #280, opened 2018-03-23
[WIP] Add cgal

massich:

PR #517, opened 2022-08-02
MAINT: Test cgal

larsoner:
Closes #280
Closes #285
Closes #286

@agramfort do we need `cgal` support? If so, this will hopefully add a test for it, assuming the 22.04 binaries are new enough and working.

If they don't work, they didn't work ~4 years ago either, given the difficulty of supporting it perhaps we should drop it? (And even if it does work, maybe we should remove it if it's not used much by people?)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -40,6 +40,8 @@ jobs:
         #
         # MKL is tested on all OSes.
         #
+        # VTK and CGAL are only tested on one Ubuntu build.
+        #
         # Someday we could use OpenBLAS for Windows once vcpkg sorts things out
         # (https://github.com/microsoft/vcpkg/issues/25176), but it's unclear
         # if this will really help anything. NumPy/SciPy's OpenBLAS (which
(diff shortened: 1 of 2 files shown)


papadop:
Yes CGAL is broken due to changes in the CGAL interface.
I intend to revive those tool later...
I need to get some help from CGAL specialists.

papadop:
After that eventually, we can add some CGAL tests. But this is not a priority.

larsoner:
Okay, we can leave this open for now then. Feel free to cherry-pick in another branch if you start working on it, or commit directly to my branch. If it gets out of date we can close and revisit later, too!

papadop:
Indeed the licence is a problem. The only way I see it is to release the CGAL part with a GPL licence.... I'll ask our legal team here if there is a better way of doing things.

And by the way, I'm almost done with the CGAL modernization stuff. It gives some tools that can be useful for the community. It would be a pity to drop all of this. Another, possibility would be to split the work in a different repository, use OpenMEEG as a library and put a GPL licence on it (or whatever solution the "lawyers" may find). I'm a bit surprised by the GPL strategy of CGAL, but that's what we have to play with unfortunately.

Last, but not least, I know fairly well (part of) the CGAL team (there are one floor below my office and some of them are friends), so I can certainly go and ask them for details. The legal team consequently also knows perfectly well the state of all of this, so I'm sure they will find a proper solution.

larsoner:
Okay! I'll close but we can reopen if the solution ends up being a project split.

PR #519, opened 2022-08-03
Remove all linux warnings

papadop:
Well without activating some optional features such as vtk...
Most of them were unused variables or parameters, initialization ordering of class members.
But it also pointed out two genuine bugs:
- Some unfinished work on the block matrix API (function to get the conductivity factors for each block).
- Some bug at the end of CorticalMat2 which was supposed multiplying the matrix by gamma, but instead just multiplying the diagonal multiple times. This function is exposed in om_assemble, but it seems that the "gamma" variant is not tested by the testsuite. @eolivi do you remember anything about this ? I believe you did that work initially...

All tests are apssing on fedora 36. Let's see what the checkers will say.


papadop:
I forgot to add. One type of warning was with the option handling in assemble.cpp. Thus I homogenized option handling in all the apps....

papadop:
Hum older gcc seems to have a problem...

agramfort (commented):
@papadop thanks 🙏 

is there a chance you can add tests for the bug fixes so do not recreate the same bugs in the future?

also I think it would be neat to maintain a changelog with bugfixes between releases

larsoner:
> also I think it would be neat to maintain a changelog with bugfixes between releases

If we title PRs well enough, the GitHub Releases automatic changelog is maybe good enough

larsoner:
I pushed a two commits: one to enable WERROR_OPT in all OSes in `build_and_test.yml`, and another to do it in `cibuildwheel`. Feel free to `git pull origin Warnings` and you should get them @papadop .

I'm pretty sure we want the `build_and_test.yml` one. I'm less sure about the `cibuildwheel.yml` one -- the gcc there is older (I think?) and I'm guessing that any additional warnings we get there are not as accurate/relevant as the ones from the newer GCC versions in the Ubuntu runners in `build_and_test.yml`.

larsoner:
(also feel free to force push over these if you had commits or changes pending, it's trivial to re-add these changes once this PR seems ready to go)

papadop:
> @papadop thanks pray
> 
> is there a chance you can add tests for the bug fixes so do not recreate the same bugs in the future?

Well, it is a part of OM which I do not know well. This is why I asked @eolivi for some input. Then indeed we can provide some tests.

> also I think it would be neat to maintain a changelog with bugfixes between releases

Or a Release notes file (though this usually generates tons of git conflicts)...

papadop:
But for now, I introduced more problems than I solved. I need to get a functional ubuntu 18.04 to understand the problem.
Almost there, but of course the cmake was too old (3.10). It is even too old to build cmake itself.... I do not know how you @larsoner managed to get a modern cmake in the testers....

larsoner:
> I do not know how you @larsoner managed to get a modern cmake in the testers....

In `cibuildwheel` I do `pip install cmake` and it installs 3.23.x I think

papadop:
Indeed old compilers give even more warnings. Which do not seem to be spurious actually....

larsoner:
Okay I made it so it no longer tries to enable WError on cibuildwheel or Ubuntu18-04. @papadop do you think it's possible to fix the warnings on macOS, Windows, and Ubuntu 20.04/22.04? Or is this what you tried to do already but couldn't get to work?

larsoner:
On Ubuntu 20.04 example it's:
```
In file included from /home/runner/work/openmeeg/openmeeg/OpenMEEG/include/triangle.h:45,
                 from /home/runner/work/openmeeg/openmeeg/OpenMEEG/include/mesh.h:50,
                 from /home/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:45:
/home/runner/work/openmeeg/openmeeg/OpenMEEG/include/vect3.h: In member function ‘void OpenMEEG::Mesh::smooth(const double&, const unsigned int&)’:
/home/runner/work/openmeeg/openmeeg/OpenMEEG/include/vect3.h:106:64: error: ‘newpt.OpenMEEG::Vect3::m[1]’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  106 |         void operator+=(const Vect3& v)  { m[0] += v.x(); m[1] += v.y(); m[2] += v.z(); }
      |                                                           ~~~~~^~~~~~~~
/home/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:186:24: note: ‘newpt.OpenMEEG::Vect3::m[1]’ was declared here
  186 |                 Vertex newpt;
      |                        ^~~~~
In file included from /home/runner/work/openmeeg/openmeeg/OpenMEEG/include/triangle.h:45,
                 from /home/runner/work/openmeeg/openmeeg/OpenMEEG/include/mesh.h:50,
                 from /home/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:45:
/home/runner/work/openmeeg/openmeeg/OpenMEEG/include/vect3.h:106:79: error: ‘newpt.OpenMEEG::Vect3::m[2]’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
make[2]: Leaving directory '/home/runner/work/openmeeg/openmeeg/build'
  106 |         void operator+=(const Vect3& v)  { m[0] += v.x(); m[1] += v.y(); m[2] += v.z(); }
      |                                                                          ~~~~~^~~~~~~~
/home/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:186:24: note: ‘newpt.OpenMEEG::Vect3::m[2]’ was declared here
  186 |                 Vertex newpt;
      |                        ^~~~~
In file included from /home/runner/work/openmeeg/openmeeg/OpenMEEG/include/triangle.h:45,
make[1]: Leaving directory '/home/runner/work/openmeeg/openmeeg/build'
                 from /home/runner/work/openmeeg/openmeeg/OpenMEEG/include/mesh.h:50,
                 from /home/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:45:
/home/runner/work/openmeeg/openmeeg/OpenMEEG/include/vect3.h:106:49: error: ‘newpt.OpenMEEG::Vect3::m[0]’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  106 |         void operator+=(const Vect3& v)  { m[0] += v.x(); m[1] += v.y(); m[2] += v.z(); }
      |                                            ~~~~~^~~~~~~~
/home/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:186:24: note: ‘newpt.OpenMEEG::Vect3::m[0]’ was declared here
  186 |                 Vertex newpt;
      |                        ^~~~~
cc1plus: all warnings being treated as errors
```

papadop:
Well. I had to setup an ubuntu 18.04 machine. And on it I fixed all the warnings that appeared, Let's see if know all tests are passing and then we can try again to enable WERROR everywhere. It seemed that on windows and MacOS, it was already passing... I did no tests on these OSes, but the tests were passing earlier today so that I somewhat confident.

larsoner:
Looks like an error on `cibuildwheel` now (where `WError` is not enabled actually):
```
    openmeeg/openmeeg_wrap.cpp: In function ‘PyObject* OpenMEEG_Vertex_array(const OpenMEEG::Vertex*)’:
    openmeeg/openmeeg_wrap.cpp:6356:41: error: lvalue required as unary ‘&’ operand
     6356 |         const double* data = &((self)->x());
          |                               ~~~~~~~~~~^~~
```
This is in the swig step

papadop:
Yeah, I think we need to remove the -Werror for swig generated code. I do not feel like trying to fix swig :-) ....
On the other hand, I think that turning warnings into errors was not effective before on MacOS and Windows as I now see a ton of warnings (errors) that were not showing before.... Indeed, I changed the logic for turning the flags, but I do not see why it would work while the previous did not .... Any idea @larsoner ?

larsoner:
> Any idea @larsoner ?

I think the logic before / on `main` was broken, it was:
```
        if [[ "${{matrix.os}}" != 'ubuntu*' ]]; then
          WERROR_OPT="-DENABLE_WERROR=OFF"
        else
          WERROR_OPT="-DENABLE_WERROR=ON"
```
So it was actually only enabled on Ubuntu before (?).

But I'm not sure how it was even enabled there, since it seems like we should have had failures like the one we have now for Ubuntu 22.04:

https://github.com/openmeeg/openmeeg/runs/7669089567?check_suite_focus=true

In any case, +1 for fixing warnings and having it only run on Linux for now.

larsoner:
I fixed the `-L` in cibuildwheel in https://github.com/openmeeg/openmeeg/pull/520 by using LINKER_OPT

papadop:
OK I limited Werror to linux (ubuntu-*). On MacOS after the linker stuff, I had tons of warning with lapack and _Complex.
On windows I have also many failures, some of which might be interesting to look at.... But I want to proceed on other things for now (it was supposed to be a 30min work...).

larsoner:
Error on 22.04:
```
In copy constructor ‘constexpr OpenMEEG::LinOpInfo::LinOpInfo(const OpenMEEG::LinOpInfo&)’,
    inlined from ‘virtual OpenMEEG::LinOpInfo OpenMEEG::maths::MatlabIO::info(std::ifstream&) const’ at /home/runner/work/openmeeg/openmeeg/OpenMEEGMaths/include/MatlabIO.H:169:88:
/home/runner/work/openmeeg/openmeeg/OpenMEEGMaths/include/linop.h:67:32: error: ‘linop.OpenMEEG::LinOpInfo::num_cols’ may be used uninitialized [-Werror=maybe-uninitialized]
   67 |     class OPENMEEGMATHS_EXPORT LinOpInfo {
      |                                ^~~~~~~~~
In file included from /home/runner/work/openmeeg/openmeeg/OpenMEEGMaths/src/MatlabIO.C:40:
/home/runner/work/openmeeg/openmeeg/OpenMEEGMaths/include/MatlabIO.H: In member function ‘virtual OpenMEEG::LinOpInfo OpenMEEG::maths::MatlabIO::info(std::ifstream&) const’:
/home/runner/work/openmeeg/openmeeg/OpenMEEGMaths/include/MatlabIO.H:166:27: note: ‘linop’ declared here
  166 |                 LinOpInfo linop;
      |                           ^~~~~
cc1plus: all warnings being treated as errors
```
If this one is too tough to fix we could restrict to `ubuntu-20.04` for now, but it would be great to cover 22.04 as well since that's the newest gcc we test with I think

OpenMEEG/src/assembleHeadMat.cpp
```diff
@@ -136,6 +136,7 @@ namespace OpenMEEG {
         }
     }
 
+    #if 0
```
larsoner, line 139: Should we remove this code then?

larsoner (commented):
Just one possible cruft I noticed, otherwise LGTM +1 for merge!

papadop:
Finally all tests are passing.
@larsoner : Do we want to add one ubuntu+clang tester with -Werror ?

OpenMEEG/src/assembleHeadMat.cpp
papadop: Well it is work in progress, I intend to finish soon. So if we can keep it in, I prefer.

larsoner:
> Do we want to add one ubuntu+clang tester with -Werror

I'd rather get one of the macOS builds working with `-Werror`. That way we don't have to add a new job, and are making sure it works on a config people will use (macOS + default compiler) rather than a less common/unused one (Linux + non-default compiler). Do you think it's possible/reasonable to do this?

PR #520, opened 2022-08-03
ENH: Add OpenMP to macOS wheels

### LINKED ISSUE #513: ENH: Use OpenMP in wheels for macOS
Currently we don't because clang doesn't support it by default. We shouldn't use `brew` like in our CIs (currently) because `brew` builds a machine-optimized version (no way to set a low MACOSX_DEPLOYMENT_TARGET), so instead we can `brew install llvm` and set it to use `llvm-clang` for compilation in `cibuildwheel`. This *should* give use OpenMP support on our macOS wheels.

larsoner:
Let's see if this works.

I'm trying to clean up the CXX flags as well by passing the linker ones only during the linking step (one fewer warning during build).

WIP to remember to restore the `build_and_test.yml` job

Closes #513


larsoner:
@agramfort feel free to review/merge if happy

agramfort (approved):

PR #523, opened 2022-08-04
Remove the boost fallback for share_ptr of arrays.

papadop:
It seems that nowadays all platforms we test have the shared_ptr extension support introduced with C++-17.
This PR tries to remove the boost fallback that we had for this (this was mainly for older MacOS compilers, as far as I can tell from the logs).


cmake/CompilerOptions.cmake
```diff
@@ -12,19 +12,6 @@ if(CMAKE_CXX_COMPILER_LOADED)
              ${CMAKE_CXX_COMPILER_VERSION}")
 endif()
 
-# Workaround the missing support of shared_ptr for arrays.
-
-check_cxx_feature(HAVE_SHARED_PTR_ARRAY_EXTENSION
-                  shared_ptr_array_extension.cpp
-                  "has C++17 extension of shared_ptr for arrays")
```
larsoner, line 19: I vaguely recall that this test failed on macOS without `boost`. Do you know if the test was just broken/incorrect somehow?

You should be able to remove the file `./cmake/utils/cxx_tests/shared_ptr_array_extension.cpp` now I think

larsoner: ... and actually I guess a lot of those could maybe be deleted now that we trimmed our set of `check_cxx_feature` tests

papadop: > I vaguely recall that this test failed on macOS without `boost`. Do you know if the test was just
> broken/incorrect somehow?

Yes it was circa 2019 for MacOS on travis, which were probably lagging behind in terms of OS release.... 
That's when I added the workaround.
I checked the configure section of all builds (except wheels which I still do not understand quite well and for which I do not see where is the configuration step) and for all of them the check_cxx_feature passed, i.e. all compilers we use have the feature now.
 
> You should be able to remove the file `./cmake/utils/cxx_tests/shared_ptr_array_extension.cpp` now I think
Indeed.

papadop: > ... and actually I guess a lot of those could maybe be deleted now that we trimmed our set of `check_cxx_feature` tests

Indeed too.

papadop:
All tests are passing (so the wheel do not need boost too). I will remove the unnecessary feature files.
I leave it to you @agramfort, @larsoner to decide. You know MacOS much better than I do.

cmake/CompilerOptions.cmake
larsoner: Click details on the macos cibuildwheel run to get to 

https://github.com/openmeeg/openmeeg/runs/7670824167?check_suite_focus=true

Then go to the "run cibuildwheel" step, then the library build you can see if you hit/uncollapse "› Running before_all...":

https://github.com/openmeeg/openmeeg/runs/7670824167?check_suite_focus=true#step:4:450

It doesn't say anything about shared pointers but I guess it works! The wheels are tested using `pytest`.

larsoner (approved):
Hooray red lines! 👍 

I clicked the "merge when green button", thanks in advance @papadop !

papadop:
Just for the record, I discovered that several of the feature tests were still in autoconf syntax and were never converted to cmake !

agramfort (approved):
❤️ red lines 👍

PR #524, opened 2022-08-04
MAINT: Test ATLAS

larsoner:
Let's see if I got this dependency right.

@papadop feel free to separately remove veclib. With these two changes I think we'll be consistent in terms of what our code has and what we say we support

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -162,8 +162,9 @@ build OpenMEEG on GitHub Actions: `.github/workflows/build_and_test.yml <https:/
 
 Supported Blas/Lapack Implementations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-We support `OpenBLAS <http://www.openblas.net/>`_ and
-`Intel MKL <http://software.intel.com/en-us/intel-mkl/>`_ on Linux, macOS, and Windows.
+We support `OpenBLAS <http://www.openblas.net/>`_,
+`Intel MKL <http://software.intel.com/en-us/intel-mkl/>`_,
+and `ATLAS <http://math-atlas.sourceforge.net/>`_ on Linux, macOS, and Windows.
 
 Using OpenMEEG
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -162,8 +162,9 @@ build OpenMEEG on GitHub Actions: `.github/workflows/build_and_test.yml <https:/
 
 Supported Blas/Lapack Implementations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-We support `OpenBLAS <http://www.openblas.net/>`_ and
-`Intel MKL <http://software.intel.com/en-us/intel-mkl/>`_ on Linux, macOS, and Windows.
+We support `OpenBLAS <http://www.openblas.net/>`_,
+`Intel MKL <http://software.intel.com/en-us/intel-mkl/>`_,
+and `ATLAS <http://math-atlas.sourceforge.net/>`_ on Linux, macOS, and Windows.
```
agramfort, line 167: Atlas on something else than Linux ?

larsoner: I'll clarify

larsoner:
@papadop when you get a chance, if we want to continue supporting ATLAS can you look at / fix https://github.com/openmeeg/openmeeg/runs/7677592445?check_suite_focus=true ? I think the right packages are installed but the `NOT LAPACKE_WORKS` code that you wrote doesn't appear to work for linking :(

larsoner:
Looks like the Generic/Netlib BLAS/LAPACK has the same problem:

https://github.com/openmeeg/openmeeg/runs/7677592541?check_suite_focus=true

So hopefully the same solution!

papadop:
If it's too complicated, let's forget about ATLAS. From the https://github.com/math-atlas/math-atlas, it looks like it is not maintained since almost 3 years....

papadop:
I will have a look.

PR #525, opened 2022-08-04
Remove Veclib support.

papadop:
The title says it all.


larsoner:
Thanks in advance @papadop !

PR #528, opened 2022-08-05
Rename conductivity_difference into conductivity_jump which is more m…

papadop:
This started by the simple renaming of the two methods:
- conductivity_difference -> conductivity_jump
- oriented -> relative_orientation
which is more meaningful and closer to mathematics.

Then I moved two methods into the .h for better readability (and inlining).
Finally, I added a fast path in the case of two identical meshes in which case the relative_orientation is necessarily equal to 1.


papadop:
All tests are passing. Merging.

PR #529, opened 2022-08-05
MAINT: Restore skipped tests

larsoner:
I don't expect this to work, but locally on my M1 machine at least things passed, so I'm curious to see which CIs fail.


larsoner:
Awesome, I guess the wrapping was a transient problem that has been fixed! In it goes

PR #530, opened 2022-08-05
ENH: Add clang-format

### LINKED ISSUE #444: MAINT: Use clang-format
We can configure this. See

https://github.com/InsightSoftwareConsortium/ITK/blob/master/.clang-format

For an example

I would not add this as commit hook or anything that requires to have clang installed but we should add this to CIs

_Originally posted by @agramfort in https://github.com/openmeeg/openmeeg/issues/392#issuecomment-549017813 but edited by @larsoner_

larsoner:
Let's see if this works.

I don't think in the end we'd want the auto-commit on PR, just on push. That way people can work on PRs in any way they want, then once the commit is merged, `clang-format` fixes the code and pushes a commit.

Closes #444

- [x] Push first commit that works, i.e., CIs green except for the clang-format checker being red: c8de1e4
- [ ] Push second commit with `[actions format]` that runs the formatter, see CIs green: 
- [ ] Use rebase+merge (*not* squash+merge) to merge PR

Separate PR:

- [ ] Follow https://github.com/numpy/numpydoc/pull/394 to skip the autoformat commit in `blame`

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,1 @@
+./OpenMEEGMaths/include/BlasLapackImplementations/complex_msvc_vendor.h
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -1,0 +1,22 @@
+name: clang-format
+on: [push, pull_request]
+jobs:
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    # https://github.com/marketplace/actions/clang-format-lint
+    - uses: DoozyX/clang-format-lint-action@v0.14
+      with:
+        source: '.'
+        clangFormatVersion: 13
+        inplace: ${{ contains(github.event.head_commit.message, '[actions format]') }}
+    - uses: EndBug/add-and-commit@v4
+      with:
+        author_name: Clang Robot
+        author_email: robot@example.com
+        message: 'Committing clang-format changes'
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      if: ${{ contains(github.event.head_commit.message, '[actions format]') }}
--- a/OpenMEEG/include/Triangle_triangle_intersection.h
+++ b/OpenMEEG/include/Triangle_triangle_intersection.h
@@ -47,6 +47,7 @@
 /*****************************************************************************/
 
 #include <cmath>
+#include <triangle.h>
 
 namespace OpenMEEG {
 
--- a/OpenMEEG/include/assemble.h
+++ b/OpenMEEG/include/assemble.h
@@ -56,9 +56,9 @@ knowledge of the CeCILL-B license and that you accept its terms.
 namespace OpenMEEG {
 
     // For ADAPT_LHS change the 0 in Integrator below into 10
-    // It would be nice to define some constant integrators for the default values but swig does not like them. 
+    // It would be nice to define some constant integrators for the default values but swig does not like them.
 
-    OPENMEEG_EXPORT SymMatrix HeadMat(const Geometry& geo,const Integrator& integrator=Integrator(3,0,0.005));
+    OPENMEEG_EXPORT SymMatrix HeadMat(const Geometry& geo,const Integrator& integrator=Integrator(3, 0, 0.005));
     OPENMEEG_EXPORT Matrix SurfSourceMat(const Geometry& geo,Mesh& sources,const Integrator& integrator=Integrator(3,0,0.005));
 
     OPENMEEG_EXPORT Matrix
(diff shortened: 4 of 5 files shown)


papadop:
Seeing the result, I definitely do not want some of the formatting....
- tab at 2 -> no, the current 4 is more readable (and I do not know who decided that 2 is a good standard, long time it was 8, which is too high, 4 is a good compromise for me).
- type& x -> type &x -> no (I know the rationale of it, but it is just ugly and poorly readable). The rationale is for when there are multiple variables defined with one type, which I prefer to forbid for pointers and references. This allows a nicer (in my view) syntax "TYPE var;" (easier to parse without having to navigate in the line).

Honestly, I largely prefer doing the formatting by hand. And as I am the main pure C++ "worker", I'd prefer not doing a huge formatting like this. Now, if the checker can attract attention on some specific bad constructs we want to avoid (e.g. avoid multiples variable pointer/reference definitions with a single type), or can enforce some very simple things (like a space after if/for/while) that would be great. But I'd clearly prefer that we choose what we want to do and introduce things step by step
(eventually doing the things by hand when the tester shows them). 

If this can be done as a first step, I'm all for it. 
But, I'm not in favor of applying a "standard" style blindly....

larsoner:
> Honestly, I largely prefer doing the formatting by hand... I'm not in favor of applying a "standard" style blindly....

The problem is that once multiple people start working, styles collide, things get harder to read, it's harder to translate knowledge across codebases (because each has its own style), etc. I think this is part of the motivation behind projects doing this sort of thing, and of multiple projects trying to adopt the same style as much as possible. In some sense this is perhaps sometimes done "blindly", but I think it could be more kindly stated as "collectively, accepting compromises".

FWIW to me the cost of learning a new format -- or really in practice, making my IDE enforce it for me! -- has been worth it on other projects at least. NumPy and other scientific Python projects (e.g., scikit-learn) have adopted Black for example, and we plan to do it in MNE-Python soon, too. Personally I don't like Black's style choices, but I buy the community-standard-adoption motivation enough to live with it at least. 🤷 

> I'd clearly prefer that we choose what we want to do and introduce things step by step

I'm not sure why step-by-step would be better. This is just more commits to ignore, more things to discuss, etc.

It sounds like you're essentially advocating for a "blocklist" approach, where we start by allowing *any* style (current practice), then selectively start adding restrictions (blocking specific style choices). This is not great for standardization as it keeps things as open as possible. It's better for standardization to follow the "allowlist" approach, where we adopt some broadly accepted coding guideline as much as we can tolerate it, i.e., making as few exceptions to it as possible. I think this approach was what the four other devs meant in terms of being in favor of `clang-format` in https://github.com/openmeeg/openmeeg/issues/444 and #392 at least.

Could we try iterating a bit by me first adding your two suggested exceptions (tabs=4 instead of 2; handling of `type& x`) to a file and then reformatting the codebase, then you see if you think you can live with the remaining `clang-format` choices, and if not, I add another exception? I'm optimistic that after only 2 or 3 iterations of this add-exception-reformat cycle we can converge to something usable, and not be too far from plain `clang-format`. And I don't think it would be super time consuming to try at least!

> eventually doing the things by hand when the tester shows them

Sure, we don't have to have a `clang-format` auto-commit corrections or anything. We can make CI just turn red and make users fix things, is is already the case in c8de1e417c2d6b1ecb6fe3fe8cb369543393c40a (and is what we've done for years in MNE-Python, too).

There are lots of options for making this easier for devs, too. In practice I would personally just set my IDE (VSCode) to `clang-format` for me. We could easily also add a precommit hook that automatically formats code for people, or just yells at them before they push so that they have to do it manually. For now I've added a trick where if you add `[actions format]` to your commit message it will reformat for you and push a commit, but this is already optional, too.

agramfort:
I am fully aligned with Eric. 

I learnt to accept black for python and it makes it clearer for new contributors what the community endorses. 

It’s the power of a standard. It avoids annoying discussion and allows to collectively agree. 

The more we add exceptions to the standard the less the standard becomes useful

papadop:
In theory I agree with you, but let's be practical. OpenMEEG is not (and will probably never be) a project with a very broad set of contributors. It is very technical, and the interested community is more users than programmers. It is hardly comparable to projects like numpy, scikitlearns or mne python (the first two have a very wide audiance and the last one still has a wider audience than OpenMEEG, but at leat is in a scripting language which is easier for users with less programming skills).

And mostly, the style is not bad. Certainly not perfect for sure, but usually readable for most of the files.
But indeed, there might be places which I may have missed to prettyfy....

And do not think I like to do that by hand. It just that I have tried these tools, with really mixed feelings on the results. Again, I will have a new look at those since my experience is from several years back, but that's not my priority now.

agramfort:
I won't argue more. If we have some spare time I think we need to focus now
on usability in the Python world
and make sure our documentation is correct and accessible. We make changes
to the C++ code changing Python
bindings API and we barely (if any?) change any python testing file. This
really tells me that the Python bindings
are currently not really tested. I would invest my time on this (note I am
on holidays)

larsoner:
Okay closing this and the related issue for now

PR #531, opened 2022-08-05
Assert rework

papadop:
When doing something else, I made some index error and saw that assert never throws the intended exception 
(so that we get tons of messages).

This was hiding a bug in the mesh file format (from BrainVisa I guess) reader, which I attempted to correct, but I do not find the documentation for that file format (broken link in an old BrainVisa forum). So I had to improvise. I ended supposing that the reader was correct and the writer not. Checked that I could read an old brainvisa .mesh file I had on a disk. I guess that if the reader was wrong, there is zero chance that om_mesh_info would say "Mesh orientation correct (valid for closed mesh)."

If you know a reader (in mne python ?) or a documentation for that format , I'd be happy to check I made the right choice.
That also probably means that no-one is using the .mesh file format writer in our users.... But for backward compatibility, I prefer to keep the support (the reader was not buggy so people may have used it without troubles).

### DIFF
--- a/OpenMEEG/include/MeshIOs/mesh.h
+++ b/OpenMEEG/include/MeshIOs/mesh.h
@@ -62,12 +62,12 @@ namespace OpenMEEG::MeshIOs {
 
             unsigned char uc[5];
             fs.read(reinterpret_cast<char*>(uc),5); // File format
-            fs.read(reinterpret_cast<char*>(uc),4); // lbindian
+            fs.read(reinterpret_cast<char*>(uc),4); // little/big endian
             //  TODO: we should check that these values are correct.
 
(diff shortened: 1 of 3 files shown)


larsoner:
We don't read any BrainVisa files in MNE-Python. Maybe it's worth looking at https://github.com/brainvisa/anatomist-free and/or posting an issue over there (or in their mailing list?) to ask.

agramfort:
It’s probably fine to assume reader was correct 

now I don’t know who still uses this format these days

papadop:
Yeah, I had a look at anatomist-free but wasn't able to find quickly the right piece of code....
All tests are passing, Merging.

PR #533, opened 2022-08-06
MAINT: Remove PyPy test skips

larsoner:
Toward https://github.com/openmeeg/openmeeg/issues/508, let's see what fails if we remove all BAD_PYPY skips.

PR #532, opened 2022-08-06
MAINT: Restore test on MKL

larsoner:
Toward #508, let's see what fails if we remove all ~~`BAD_MSVC`~~ `BAD_MKL` skips.

Closes #508

PR #536, opened 2022-08-09
Improved Python error handling

papadop:
Calling PyErr_SetString does not interrupt the program flow, so does not generate the expected exception. The program was continuing so eventually not even generating an error or not the first one. This happened when an "extend" was calling other "extend" functions/methods. Instead, I introduced exceptions and let the swig mechanism to handle exceptions to deal with those.

The messages are also improved, so instead of and uninformative runtime error (because for some reason the PyErr_SetString message was lost), we now get:
>>> mesh1 = om.Mesh(np.array([[0.0, 0.0, 0.0]]), triangles)
  File "<stdin>", line 1, in <module>
  File "/home/papadop/compiles/OpenMEEG/OpenMEEG-ref-f36/wrapping/python/openmeeg/openmeeg.py", line 1306, in __init__
    _openmeeg.Mesh_swiginit(self, _openmeeg.new_Mesh(*args))
ValueError: Vertex index 1 of triangle 0 out of range

I converted TypeError messages into ValueError messages because the former was giving a stack trace that did not seem very meaningful ti me, but if a python expert tells me a better type, it is easy to change that....

### DIFF
--- a/wrapping/python/openmeeg/openmeeg.i
+++ b/wrapping/python/openmeeg/openmeeg.i
@@ -4,6 +4,9 @@
 
 %inline %{
 
+    #include <string>
+    #include <exception>
+
     class Error: public std::exception {
     public:
 
--- a/wrapping/python/openmeeg/openmeeg.i
+++ b/wrapping/python/openmeeg/openmeeg.i
@@ -2,12 +2,32 @@
 
 %feature("autodoc", "1");
 
+%inline %{
+
+    class Error: public std::exception {
+    public:
+
+        Error(const int t,const char* m): typ(t),mes(m) { }
+
+        int type() const { return typ; }
+
+        const std::string& message() const { return mes; }
+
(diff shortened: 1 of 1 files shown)


papadop:
Not sure this can be related to the pypy problems #532 and #533.

papadop:
All tests are passing. Merging.

larsoner:
@papadop it looks like you converted only some of the `PyErr_SetString` into `throw Error(...)`. Why not convert them all?

larsoner:
... also there are a lot of `exit(...)` in the code -- that's causing the silent exit I'm debugging in https://github.com/openmeeg/openmeeg/pull/540. I'm not sure the best way to handle this. Maybe these should all become thrown exceptions, then these should be handled at the app level?

larsoner:
... moving discussion to #541 so we can track in a proper issue

PR #543, opened 2022-08-16
Convert to std::filesystem.

papadop:
Use std::filesystem instead of our own routines where possible.

### DIFF
--- a/OpenMEEG/include/GeometryIOs/GeomFile.h
+++ b/OpenMEEG/include/GeometryIOs/GeomFile.h
@@ -42,6 +42,8 @@ knowledge of the CeCILL-B license and that you accept its terms.
 #include <map>
 #include <vector>
 #include <string>
+#include <fstream>
+#include <filesystem>
 
 #include <interface.h>
 #include <GeometryExceptions.H>
@@ -50,6 +52,8 @@ knowledge of the CeCILL-B license and that you accept its terms.
 #include <MeshIO.h>
 
 namespace OpenMEEG::GeometryIOs {
+
+    namespace fs = std::filesystem;
 
     // Read the head file description (.geom files) and load the information into a Geometry.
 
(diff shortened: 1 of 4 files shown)


larsoner:
[Windows](https://github.com/openmeeg/openmeeg/runs/7857650362?check_suite_focus=true#step:4:825):
```
    D:\a\openmeeg\openmeeg\OpenMEEG\include\GeometryIOs/GeomFile.h(133): error C2440: '<function-style-cast>': cannot convert from 'std::filesystem::path' to 'std::string' [D:\a\openmeeg\openmeeg\build\OpenMEEG\OpenMEEG.vcxproj]
```
[macOS](https://github.com/openmeeg/openmeeg/runs/7857650218?check_suite_focus=true#step:4:733):
```
  In file included from /Users/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:46:
  In file included from /Users/runner/work/openmeeg/openmeeg/OpenMEEG/include/MeshIO.h:49:
  /Users/runner/work/openmeeg/openmeeg/OpenMEEG/include/filenames.h:53:48: error: '~path' is unavailable: introduced in macOS 10.15
          const std::string ext = fs::path(name).extension();
                                                 ^
  /Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/filesystem:961:3: note: '~path' has been explicitly marked unavailable here
    ~path() = default;
    ^
  In file included from /Users/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:46:
  In file included from /Users/runner/work/openmeeg/openmeeg/OpenMEEG/include/MeshIO.h:49:
  /Users/runner/work/openmeeg/openmeeg/OpenMEEG/include/filenames.h:53:27: error: 'operator basic_string' is unavailable: introduced in macOS 10.15
          const std::string ext = fs::path(name).extension();
```
[Ubuntu 18.04](https://github.com/openmeeg/openmeeg/runs/7857651841?check_suite_focus=true#step:16:150):
```
In file included from /home/runner/work/openmeeg/openmeeg/OpenMEEG/include/MeshIO.h:49:0,
                 from /home/runner/work/openmeeg/openmeeg/OpenMEEG/src/mesh.cpp:46:
/home/runner/work/openmeeg/openmeeg/OpenMEEG/include/filenames.h:43:10: fatal error: filesystem: No such file or directory
 #include <filesystem>
          ^~~~~~~~~~~~
```

papadop:
OK. Too bad these old compilers do not implement that. It is C++ 17. I checked that windows has it, and thought all others will have it if it's in windows...

Let forget about that change for now... I'll keep the branch around for a future. I do not want to re-introduce a boost dependency for this....

larsoner:
We might "just" be able to install GCC 9 on Ubuntu 18.04 and have it work there, I can push some commits to do that if you want. We say we're c++17 and `<filesystem>` is part of that so it seems reasonable to require it. `cibuildwheel` even worked for Linux so I think it's safe enough...

Do you think the Windows and macOS errors are fixable, though?

larsoner:
... or really we can just bump 18.04 to 20.04 since it's where c++17 is "properly supported". Seems acceptable to me. We can trust `cibuildwheel` to run the "old setup" compile

PR #545, opened 2022-08-16
Remove no longer useful pragmas.

papadop:
I noticed that some of the windows pragmas are related to old stuff that was removed years ago (exception specifications, ...). Trying to remove some of those (Removing even more WIN32 material)....


larsoner:
These mingw64 failures seem unrelated. I'll see if I can fix them separately

larsoner:
Thanks @papadop !

PR #548, opened 2022-08-16
MAINT: Remove mingw64

### LINKED ISSUE #547: build (windows-2019, mingw64) failure
On several MR I have this build (windows-2019, mingw64) that fails in the configure step on python 3 detection.
I do not think this is related to the MR. Is this a transitory bug, or did we manage to break the builders without noticing it ?

larsoner:
Closes #547


larsoner:
I think for now we should just get rid of the `mingw64` job. I'm not sure we really need to support it -- I started out using it because it worked way better than MSVC. But given that MSVC now works and it's what we need to use for wheels (and on conda-forge) anyway, I think we can drop it for now. I'll comment it out and we can think about bringing it back if someone needs it.

larsoner:
From a quick chat with @agramfort he's okay with dropping mingw64 so I'll merge

PR #552, opened 2022-08-16
BUG: Convert exit to exceptions

### LINKED ISSUE #541: MAINT: Unify error handling
... also there are a lot of `exit(...)` in the code -- that's causing the silent exit I'm debugging in https://github.com/openmeeg/openmeeg/pull/540. I'm not sure the best way to handle this. Maybe these should all become thrown exceptions, then these should be handled at the app level?

@papadop it looks like you converted only some of the `PyErr_SetString` [in #5365] into `throw Error(...)`. Why not convert them all?

_Originally posted by @larsoner in https://github.com/openmeeg/openmeeg/issues/536#issuecomment-1215638783_

_Originally posted by @larsoner in https://github.com/openmeeg/openmeeg/issues/536#issuecomment-1215679749_

### LINKED ISSUE #551: MAINT: Remove French language conditionals?
Looking into exception handling in `FileExceptions` a bit, @agramfort added 15 years ago some conditionals for `LANGUAGE`. Can we remove these?

larsoner:
@papadop I started working through `git grep "exit("` and stopped before I got too far to check in. Does this seem on the right track to you? If so, I'm happy to continue. I don't plan to mess with `commandline.h`.

Also, I propose we rename `GeometryExceptions.H` to `Exceptions.H` since it already contains some other exceptions, and it's nice to have them all in one header with a more generic name. We could alternatively create more `*Exceptions.H` files, but I'm not sure there's much value there.

- [x] remove all non-command-line `exit()`s
- [x] ~~Migrate to Exceptions.H~~
- [x] fix Python RuntimeError exception that should be IOError
- [x] Remove french language "localization"

Closes #551
Closes #541


larsoner:
... we should maybe also put the `FileExceptions.H` values in `Exceptions.H` and use the same naming scheme like `OpenMEEG::ReadError` rather than `read_error`. This would help with unification, and with this comment that I left over in https://github.com/openmeeg/openmeeg/pull/553 for MeshIO.h:
```
            if (!fs.is_open()) {
                std::ostringstream ost;
                ost << "Error opening " << name() << " file: " << fname << " for reading." << std::endl;
                // TODO this should use read_error from FileExceptions.H, then we should look for io_error in openmeeg.i and reraise as SWIG_IOError
                throw std::invalid_argument(ost.str());
```

papadop:
This is completely on-track... Thanks.

papadop:
> Also, I propose we rename `GeometryExceptions.H` to `Exceptions.H` since it already contains some other exceptions, and it's nice to have them all in one header with a more generic name. We could alternatively create more `*Exceptions.H` files, but I'm not sure there's much value there.

This is OK with me. A minor nit is that the two files do not follow exactly the same logic. FileExceptions.H do not provide a return code. But this is actually good otherwise we might have had a trouble with conflicting return codes. That being said, it seems that FileExceptions.H is imported from another code of mine (and then changed of course), probably like most of the files ending with .H.  It is used only in DataTag.H and only one exception is used (read_error). So there is no real value in keeping that file, just migrate the read_error exception (and do not put it in the std namespace) and use one single Exceptions.H file.

larsoner:
> So there is no real value in keeping that file, just migrate the read_error exception (and do not put it in the std namespace) and use one single Exceptions.H file.

Okay will do!

larsoner:
In the end I kept `OpenMEEG/include/GeometryExceptions.H` named that way because we already have `OpenMEEGMaths/include/Exceptions.H`. Eventually these could maybe be refactored -- there is a lot of code dup / similarity between the `OpenMEEGMaths` version and the `OpenMEEG` one -- but I'll leave that for another PR. After this at least the Python exceptions are decent, and the `exit`s are in `commandline.h`, `apps/`, and `tests/`, cmake options, and utils only:

<details>
<summary>git grep "exit("</summary>

```
$ git grep "exit("
OpenMEEG/include/commandline.h:                exit(1);
OpenMEEG/include/commandline.h:                        exit(1);
OpenMEEG/include/commandline.h:            exit(1);
OpenMEEGMaths/tests/full.cpp:        exit(1);
OpenMEEGMaths/tests/full.cpp:                    exit(1);
OpenMEEGMaths/tests/full.cpp:                    exit(1);
OpenMEEGMaths/tests/full.cpp:        exit(1);
OpenMEEGMaths/tests/full.cpp:        exit(1);
OpenMEEGMaths/tests/sparse.cpp:        exit(1);
OpenMEEGMaths/tests/test_mat_files_io.cpp:        exit(1);
OpenMEEGMaths/tests/test_mat_files_io.cpp:        exit(1);
OpenMEEGMaths/tests/vector.cpp:        exit(1);
apps/assemble.cpp:            exit(1);
apps/assemble.cpp:            exit(1);
apps/assemble.cpp:            exit(1);
apps/assemble.cpp:        exit(1);
apps/gain.cpp:    exit(1);
apps/gain.cpp:        exit(1);
apps/minverser.cpp:    exit(0);
apps/tools/make_nerve.cpp:            exit(1);
apps/tools/make_nerve.cpp:            exit(1);
apps/tools/make_nerve.cpp:    exit(0);
apps/tools/matrix_info.cpp:        exit(1);
build_tools/upload_package_gforge.py:        sys.exit(0)
cmake/utils/matioVersion.c:    exit(0);
tests/test_compare_matrix.cpp:        exit(1);
tests/test_compare_matrix.cpp:        exit(1);
tests/test_compare_matrix.cpp:            exit(1);
tests/test_compare_matrix.cpp:            exit(1);
tests/test_compare_matrix.cpp:        exit(1);
tests/test_compare_matrix.cpp:        exit(1);
tests/test_compare_matrix.cpp:            exit(1);
tests/test_compare_matrix.cpp:            exit(1);
tests/test_compare_matrix.cpp:            exit(1);
tests/test_compare_matrix.cpp:            exit(1);
tests/test_mesh_ios.cpp:        exit(1);
tests/test_validationEIT.cpp:    exit(0);
```

</details>

Ready for review/merge from my end!

papadop:
This is great. I have a small issue with the geometry.cpp load/save change, which I think looses some information. The IO routines, I think, have no knowledge of the filename, so the exception they throw cannot contain that information. By just letting the original exception to flow to the main app, I think we loose the filename information which is a big help for the user to know where the problem occured. So we need to catch the exception and rethrow smthg that also has the filename information so that a complete message will be printed. I'll try to see if I can do that if you do not beat me.

I have not looked closely at all the changes, but besides this I think the change is OK. I will have a closer look later,

larsoner:
Feel free to push directly the exception improvements @papadop !

papadop:
Hum MeshIO.h file is also suspect. The original message uses name() and fname, while the new only contains name(). I need to dive in the code a little more to understand what this means.

larsoner:
I pushed a commit to merge with `main`, hopefully I did it right and it comes back green.

@papadop once you improve the error messages here I plan to merge and then push out 2.5.0 -- sound good?

papadop:
OK I pushed an enhanced (?) version.
In addition to correct the previously mentioned things, I renamed the file GeometryException.H -> OMExceptions.H, Exceptions.H -> OMMathExceptions.h and moved from throw() specification to noecept (sincxe C++ 11).

OpenMEEGMaths/include/BlasLapackImplementations/complex_msvc_vendor.h
```diff
@@ -159,112 +159,112 @@ extern "C++"
     // double complex overloads
     //
     //-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    inline _Dcomplex __CRTDECL acos(_In_ _Dcomplex _X) throw()
+    inline _Dcomplex __CRTDECL acos(_In_ _Dcomplex _X) noexcept
```
larsoner, line 162: Can you revert these? We should keep this vendored file unchanged ideally

larsoner: I did this with a quick `checkout`, will set to merge when green!

PR #553, opened 2022-08-16
MAINT: Convert PyErr to throw

larsoner:
Second part of #541. Removed all `PyErr_SetString` in favor of the `throw Error(Swig_*` approach. Let's see if things are still okay here. I changed a few types to more closely match what I think makes sense for the Python error types.

I did not change `numpy.i` since that should just be a copy/vendored file.


larsoner:
Okay this one is good to go @agramfort @papadop !

agramfort (approved):

papadop:
Thanks...  Feel free to add yourself in the python contributors as soon as the page is published ;-)

PR #554, opened 2022-08-16
MAINT: Correct code for black and flake8

larsoner:
Over in https://github.com/openmeeg/openmeeg/pull/553 I hit an error because we had no `flake8`-sort of checking going on (`pyflakes` would have caught it), so I want to add some Python style checks to CIs.

At the same time, many scientific Python groups are moving to `black` (NumPy, scikit-learn). It seems like a good time since there isn't too much Python code in the repo, and we can just standardize around it easily.

In this PR I'm:

1. changing the Python code to conform to these styles
2. adding a pre-commit hook that enables automatic `black` formatting if people want to `pre-commit install` it, but it's not required
3. adding CI checks for `black` and `flake8` compliance

This should go in after https://github.com/openmeeg/openmeeg/pull/553, hence the WIP

### DIFF
--- a/wrapping/python/openmeeg/_distributor_init.py
+++ b/wrapping/python/openmeeg/_distributor_init.py
@@ -4,20 +4,21 @@ import os
 import sys
 import platform
 
-if os.name == 'nt':
+if os.name == "nt":
     from ctypes import WinDLL
     import glob
+
     libs_path = os.path.abspath(os.path.join(os.path.dirname(__file__)))
     try:
         owd = os.getcwd()
         os.chdir(libs_path)
-        for filename in glob.glob(os.path.join(libs_path, '.libs', '*dll')):
+        for filename in glob.glob(os.path.join(libs_path, ".libs", "*dll")):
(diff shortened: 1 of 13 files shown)


larsoner:
The lint checks take 9s and 7s, respectively, so the additional CI burden is negligible here I think!

larsoner:
@agramfort @papadop let me know if you're okay with this direction and if so I'll rebase after https://github.com/openmeeg/openmeeg/pull/553, show that without reformatting the `test_geom.py` file from it `flake8` and/or `black` will complain, then I'll fix the style, then merge when green

larsoner:
Okay this PR now just does all the reformatting. I'll squash and merge then in a follow-up PR add back in the CI checks and a blame-ignore-revs listing the squashed commit

agramfort (approved):

papadop (approved):

papadop:
I do not know what is black, but I'm sure you have made the right choices. Thank you.

PR #561, opened 2022-08-19
MAINT: Use std::filesystem

### LINKED ISSUE #542: Use std::filesystem
And remove OpenMEEG/include/filenames.h
It should be available on all compiler with C++-17 support

larsoner:
Let's see if this works

Closes #542

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -327,7 +327,7 @@ jobs:
           echo "CMAKE_PREFIX_PATH=$BLAS_DIR" >> $GITHUB_ENV
         fi
         # Set our version min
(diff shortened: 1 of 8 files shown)


papadop:
I thought it was a lost cause for now at least because of lack of implementation in old versions of the compilers.....

larsoner:
`cibuildwheel` can handle it, which I think is as far back as we should reasonably need to support. The compilers there should have proper C++17 support. I'm hoping for macOS I "just" need to set the version min to 10.15...

larsoner:
This is all good with minor modifications to #543, @papadop feel free to merge if you're happy!

larsoner:
From a quick look with @agramfort this was good to go!

PR #563, opened 2022-08-20
ENH: Build apps on GH and attach to releases

### LINKED ISSUE #495: MAINT: Replace release url with GH releases

larsoner:
Closes #495 

- [x] Remove `-DENABLE_APPS=OFF` from cibuildwheel
- [x] Add `--target package` to cibuildwheel
- [x] Get static linkage to OpenBLAS working
- [x] Set CI to make a draft release even on PRs
- [x] Get upload of packages to draft release working
- [x] Restore `build_and_test.yaml` and `cibuildwheel.yml` for PRs (just disabled now to save some CI cycles)
- [x] Tick version number

This will at least get us closer. Then we need to test the packages and see if the linking works, etc.


larsoner:
Okay I *think* I got static linking working for the library itself. The problem is that then building the wheels separately somehow requires us to link to OpenBLAS separately... not sure of a great solution here. We might want a separate CI run just for building the binaries :(

larsoner:
(This might have to do with `-fvisibility` / `-fvisibility-externs-dllimport` but I'm not sure...)

larsoner:
I'm going to go ahead and merge this as an infrastructure fix. I'll test all four binaries this week and make any necessary fixes separately so that this doesn't grow out of control!

PR #564, opened 2022-08-22
ENH: Add additional tests to python API

agramfort:


larsoner:
You should be able to in principle use `pre-commit` install to get automatic `black` like in scikit-learn. But it's mostly untested from my end (did not work on Ubuntu, haven't tried on M1 yet) so feel free to fix that as well if it doesn't work :)

larsoner (dismissed):
Marking as changes requested to remind us to make `flake8` actually look for unused variables, as it would have caught some of these omissions; removing F841 from here should flag the problems:

https://github.com/openmeeg/openmeeg/blob/main/wrapping/python/setup.cfg#L2

wrapping/python/openmeeg/openmeeg.i
```diff
@@ -455,8 +455,12 @@ namespace OpenMEEG {
         dims[0] = ($self)->nlin();
         dims[1] = ($self)->ncol();
 
+        // make a copy of the data
+        double* data = new double[dims[0]*dims[1]];
+        double* start = (*($self)).data();
+        std::copy(start,start+dims[0]*dims[1],data);
```
larsoner, line 461: FYI @papadop without this @agramfort (and CIs) were getting segfaults, presumably due to some memory deallocation

papadop: That's unfortunate... This means that using python will double the amount of memory used... I must say that it has always been difficult to me to handle correctly reference counting with python. One possibility would be to increment OpenMEEG reference, but then we would need to wrap python destructor to decrease the reference counter and I do not know how to do that.

So copying is the best solution in the meantime.... But maybe we can add some comment

papadop: OK, I just found:
https://numpy.org/doc/stable/reference/c-api/data_memory.html
This seems a way to handle the above problem better.
Instead of copying, we increase the array reference count in OpenMEEG. But we also set a handler whose role would be to:
1) increase the ref counter in case of a shallow copy.
2) decrease the counter when the array is destroyed by python.

I have not looked in depth at the mentioned web page, but that seems a plausible strategy.
Have you added a test that exposed the @agramfort issue (or was it already existing) ???

papadop: To comment on myself. It's even simpler because we have python ref counting for python clones. We just need to trap the intent of python to free the array and decrease the openmeeg ref counter then...

agramfort: If we don’t do this it makes the current python tests segfault

agramfort:
Honestly for me it’s low priority 

We have to do a better job with user facing code 

Our Python API is not consistent. Sometimes you accept an array sometimes you don’t. If you screw up the shape it segfaults 

If you don’t have a Fortran layout it should raise a nice exception

And we need to test better single layer model

And constructing a Sensors object in Python is a mess…

This hurts usability a lot more than having to copy an array

See XXX in the tests for example

larsoner:
> That's unfortunate... This means that using python will double the amount of memory used...

Agreed it's less than ideal. So far at least this has been negligible in testing/use in terms of memory consumption, though.

>Our Python API is not consistent. Sometimes you accept an array sometimes you don’t. If you screw up the shape it segfaults... If you don’t have a Fortran layout it should raise a nice exception ... And we need to test better single layer model
(#577)... And constructing a Sensors object in Python is a mess… This hurts usability a lot more than having to copy an array
... See XXX in the tests for example

Agreed that these are all a higher priority. And to add to the list:

- Wrappers are still unusable on Windows: https://github.com/openmeeg/openmeeg/pull/582 / https://github.com/openmeeg/openmeeg/runs/8088313405?check_suite_focus=true
- From that issue as well, I suspect that exceptions are not handled properly in OpenMP threads as they lead to a segfault rather than a nice `RuntimeError` (that does not kill the Python terminal), as once we disable OpenMP on Windows we get such a RuntimeError instead (also from #582): https://github.com/openmeeg/openmeeg/runs/8092533939?check_suite_focus=true

I'll open an issue with all of these little bugs so we can keep track and gradually check the boxes!

PR #565, opened 2022-08-23
MAINT: Disable OpenMP in wheel builds

larsoner:
@agramfort is having segfaults with the OpenMP builds on PyPi, and on his local machine. Let's disable them for now for the wheels.

The conda-forge builds should be okay I think.

Doing a version bump to 2.5.1 and will cut a release when green+merged


agramfort (approved):

larsoner:
I'll merge this when green even though 2.5.1 isn't quite yet (we'll want #564). But getting it in should allow us to test the 2.5.1 prerelease wheels uploaded to TestPyPi

PR #568, opened 2022-08-24
BUG: Force users to specify domain_name

larsoner:
@agramfort was just hit by a bug where `DipSourceMat` was not working properly because he didn't specify that `domain_name='brain'`. This PR makes the `domain_name` a mandatory argument, which seems safer than ~~allowing it to be~~ defaulting it to an empty string.

### DIFF
--- a/apps/assemble.cpp
+++ b/apps/assemble.cpp
@@ -133,7 +133,7 @@ int main(int argc, char** argv)
         assert_non_conflicting_options(argv[0],++num_options);
 
         // Computation of distributed Surface Source Matrix for BEM Symmetric formulation
-        
+
         const Geometry geo(opt_parms[1],opt_parms[2],use_old_ordering);
         Mesh mesh_sources(opt_parms[3]);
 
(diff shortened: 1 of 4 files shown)


apps/assemble.cpp
```diff
@@ -172,8 +172,8 @@ int main(int argc, char** argv)
 
         const char* optname = opt_parms[0];
         const unsigned integration_levels = check_no_adapt(optname,{"-DipSourceMatNoAdapt", "-DSMNA", "-dsmna"}) ? 0 : 10;
-        
-        const Matrix& dsm = DipSourceMat(geo,dipoles,Integrator(3,integration_levels,0.001));
+
+        const Matrix& dsm = DipSourceMat(geo,dipoles,Integrator(3,integration_levels,0.001),domain_name);
```
larsoner, line 176: I think this was perhaps a legitimate bug that was hiding behind the default, as the `domain_name` defined above is not used otherwise (other than to emit a message)

larsoner:
I'll merge this to keep the fixes going, but @papadop if you see a better solution I'm happy to follow up!

PR #569, opened 2022-08-24
BUG: Set barrier when not reading from file

agramfort:
…om file


larsoner:
I'll merge when green -- I had to remove the `mne_sample` tests because you did not push the data @agramfort , and the `.tri` files I had did not appear to work. This could be a nice follow-up PR, though!

PR #574, opened 2022-08-26
add wrapping for SymMatrix in Python

agramfort:


larsoner:
Will merge when green, thanks in advance @agramfort !

PR #576, opened 2022-08-26
ENH : add support for one layer in make_nested_geometry function

agramfort:


larsoner:
LGTM is a false alarm, marking for merge when green. Thanks @agramfort !

PR #579, opened 2022-08-27
MAINT: Remove ruby and legacy code

### LINKED ISSUE #575: MAINT: Remove ruby wrappers
Can we remove the ruby wrappers? They aren't built or tested AFAIK

larsoner:
Closes #575

We had already removed most of the ruby wrapping stuff I think, so this isn't a huge change. It does remove some cruft from the `openmeeg.i` file as well for some proof-of-concept code. This code will continue to exist in our history if we need to find it again, no need to clutter the wrapper with it.


larsoner:
@agramfort feel free to mark for auto-merge if you're happy

agramfort (approved):

PR #582, opened 2022-08-27
BUG: Fix bugs with Geometry references in make_geometry

agramfort:


larsoner:
Okay @agramfort it looks like without OpenMP enabled we get:

https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=21557&view=logs&j=305851a9-a7bb-55db-0042-7e2b6f48aa1c&t=4d37777d-f36a-53aa-9217-6386d15dddcd

```
___________________ test_make_forward_solution_openmeeg[3] ____________________
mne\forward\tests\test_make_forward.py:269: in test_make_forward_solution_openmeeg
    bem = make_bem_solution(bem_surfaces, solver=solver)
<decorator-gen-65>:12: in make_bem_solution
    ???
mne\bem.py:397: in make_bem_solution
    _fwd_bem_openmeeg_solution(bem)
mne\bem.py:341: in _fwd_bem_openmeeg_solution
    hm = om.HeadMat(geom)
E   RuntimeError: Unknown vertex in triangle
```
So maybe this is an int32 (win) versus int64 (unix) issue or something?

One helpful step forward would probably be to improve this error message to tell us which vertex it's complaining about, in which triangle... I might open a PR for this.

larsoner:
(also my guess is that this is raised within a OpenMP thread somewhere, and that isn't handled nicely by our code in OpenMEEG, so we should also fix that at some point...)

agramfort:
@larsoner know what’s happening here ?

larsoner:
> @larsoner know what’s happening here ?

It looks like you replicated the issue, which is great!

No I'm not sure what the issue is offhand, but my guess is that it's some datatype issue, e.g., `int` in Python on Windows meaning one thing but needing to be `int64` here in order to work properly. If you read from disk this sort of thing could happen easily because the C++ code could always make `int64`, but in the wrappers we might allow people to pass `int` / `int32`, but then not handle it properly (or overflow).

larsoner:
First I did a merge with `main` to see if we get a nice RuntimeError.

Next I'll try changing this to *only* allow 64-bit signed int:

https://github.com/openmeeg/openmeeg/blob/af99d4168d39123ebbefccef8fa7b328b67a3aef/wrapping/python/openmeeg/openmeeg.i#L309-L312

Hopefully with no changes to the test it first raises an error complaining that a wrong dtype is used. Then if we modify the test to actually pass a 64-bit int, hopefully everything is green.

We could then leave the investigation for "why 32-bit int is problematic" for another day, and just live with users having to do a `.astype(np.int64)` in their code.

larsoner:
... then I'll also reenable OpenMP, since it should be safe enough by then I hope

papadop:
For the record, I doubt it is a problem with an int size. This is in the code:

```
unsigned vertex_index(const Vertex& V) const {
            for (unsigned i=0;i<3;++i)
                if (&vertex(i)==&V)
                    return i;
            throw UnknownVertex();
        }
```
which basically looks for a vertex by its address. So this is more related to an array of vectors that has been re-allocated (by the copy you added ?) than on an index pb.

larsoner:
Okay now we get the same error message [on the OpenMP build](https://github.com/openmeeg/openmeeg/runs/8096392460?check_suite_focus=true#step:20:853) as we do [on the non-OpenMP build](https://github.com/openmeeg/openmeeg/runs/8096385639?check_suite_focus=true#step:6:2893), suggesting that the `ThreadException` code I added is working:
```
  >       hm = om.HeadMat(geom)
  E       RuntimeError: Unknown vertex in triangle: -0.006347 0.060848 0.117761
```
The macOS build / `clang` is broken, which we'll have to fix. I found [this thread](https://stackoverflow.com/questions/66362932/re-throwing-exception-from-openmp-block-with-the-main-thread-with-rcpp) which seems relevant. But we'll have to figure out how to handle exceptions there properly :(

> which basically looks for a vertex by its address. So this is more related to an array of vectors that has been re-allocated (by the copy you added ?) than on an index pb.

Indeed this would make sense. Any ideas for how to track it down?

papadop:
I remove the data copy change and did the test. For me the test fails on:
assert_allclose(om.Matrix(hminv).array(), om.Matrix(hminv_inplace).array())
I added an rtol up to 1e-1 and it still fails. So this is indeed strange.
Printing the arrays seems to give the same values:
```
5: [[-14.59572605  -7.39214717  -6.91887338 ...   0.07269055   0.0441663
5:     0.06522981]
5:  [ -7.39214717 -14.59572608  -6.8016229  ...   0.08909068   0.08289895
5:     0.08456384]
5:  [ -6.91887338  -6.8016229  -14.59572598 ...  -0.55284908  -0.05477033
5:    -0.09554675]
5:  ...
5:  [  0.07269055   0.08909068  -0.55284908 ...   4.02518834  -0.08156556
5:    -0.62673895]
5:  [  0.0441663    0.08289895  -0.05477033 ...  -0.08156556   4.02518834
5:    -0.62673895]
5:  [  0.06522981   0.08456384  -0.09554675 ...  -0.62673895  -0.62673895
5:     3.29065398]]
5: E1
5: [[-14.59572605  -7.39214717  -6.91887338 ...   0.07269055   0.0441663
5:     0.06522981]
5:  [ -7.39214717 -14.59572608  -6.8016229  ...   0.08909068   0.08289895
5:     0.08456384]
5:  [ -6.91887338  -6.8016229  -14.59572598 ...  -0.55284908  -0.05477033
5:    -0.09554675]
5:  ...
5:  [  0.07269055   0.08909068  -0.55284908 ...   4.02518834  -0.08156556
5:    -0.62673895]
5:  [  0.0441663    0.08289895  -0.05477033 ...  -0.08156556   4.02518834
5:    -0.62673895]
5:  [  0.06522981   0.08456384  -0.09554675 ...  -0.62673895  -0.62673895
5:     3.29065398]]
```
But when doing slightly fancier things, I get nonsense:
```
M1 = om.Matrix(hminv).array()
M2 = om.Matrix(hminv_inplace).array()
ind = np.unravel_index(np.argmax(np.abs(M1-M2), axis=None), M1.shape)
print(ind, M1[ind], M2[ind])
print(M1)
print(M2)
```
`5: (0, 0) 6.9009368116229e-310 6.9009368116229e-310`

So I really think that it's a case where the memory has been freed by openmeeg/numpy, probably because of the function call. What is strange is that in this case the array should have been created and allocated (we do not have the choice because numpy does not have a concept of a symmetric matrix), so we should not have a problem with freed array in this 
specific case....

Ah OK, I understood. The symmetric matrix is first converted to a matrix before the array method is called.  Then the array method() is called and python thinks it can destroy the underlying  om.Matrix object... So we are in the memory reference problem I exposed above.

To test it, let's try to keep a python reference of the om.Matrix.
```
    print('E')
    M1 = om.Matrix(hminv)
    M2 = om.Matrix(hminv_inplace)
    print('2')
    assert_allclose(M1.array(), M2.array())
    print('3')
    assert_allclose(om.Matrix(hminv).array(), om.Matrix(hminv_inplace).array())
    print('F')
```
which gives:
```
5: E
5: 1
5: 2
5: 3
5: FAILED
```
So the assert after 2 passes but not the one after 3. QED, this is a case where python frees the memory associated with the Matrix structure (because in my case they are not copied).

papadop:
In conclusion, we cannot copy the data, otherwise we get mistakes at the om.Geometry level which expects the vertices to have the same addresses in memory. But with the current code, python frees the Matrix memory of temporary objects, which is wrong, so we have to deal better with the memory management as explained in the web page provided earlier. I'll try to explore that later tonight or tomoroow.

larsoner:
>  I'll try to explore that later tonight or tomoroow.

Sounds good! Feel free to push here directly. In the meantime, I think I fixed the OpenMP stuff so that it no longer segfaults on an error, all of the failures I've looked at so far are `RuntimeError`s instead. Feel free to review that bit of code and clean it up if necessary as well.

papadop:
> > I'll try to explore that later tonight or tomoroow.
> 
> Sounds good! Feel free to push here directly. In the meantime, I think I fixed the OpenMP stuff so that it no longer segfaults on an error, all of the failures I've looked at so far are `RuntimeError`s instead. Feel free to review that bit of code and clean it up if necessary as well.

Yeah. I already had a look at it. That seems great. I even found the stackoverflow post that has probably inspired it !!!
This seems a nice way even though it complexifies the code to some point, but correctness is a priority (except maybe if we can prove that this should never happen AND if there is a strong time penalty in doing this).

I'm just afraid that we may have to protect many more OMP expressions in this way.

larsoner:
> I'm just afraid that we may have to protect many more OMP expressions in this way.

I think we're okay, unless you really want one for the test, or if there is something other than `#pragma omp parallel` that we need to look for:
```shell
$ git grep "[^/]#pragma omp parallel"
OpenMEEG/include/gain.h:        #pragma omp parallel for
OpenMEEG/include/operators.h:            #pragma omp parallel for
OpenMEEG/include/operators.h:                #pragma omp parallel for
OpenMEEG/include/operators.h:                #pragma omp parallel for
OpenMEEG/include/operators.h:                #pragma omp parallel for
OpenMEEG/include/operators.h:                #pragma omp parallel for
OpenMEEG/src/assembleHeadMat.cpp:                            #pragma omp parallel for
OpenMEEG/src/operators.cpp:        #pragma omp parallel for
OpenMEEG/src/operators.cpp:        #pragma omp parallel for
OpenMEEG/src/operators.cpp:        #pragma omp parallel for
cmake/utils/cxx_tests/openmp_support.cpp:    #pragma omp parallel for
larsoner@bunk:~/python/openmeeg$ git grep "ThreadException e"
OpenMEEG/include/gain.h:        ThreadException e;
OpenMEEG/include/operators.h:            ThreadException e;
OpenMEEG/include/operators.h:            ThreadException e;
OpenMEEG/include/operators.h:            ThreadException e;
OpenMEEG/include/operators.h:            ThreadException e;
OpenMEEG/include/operators.h:            ThreadException e;
OpenMEEG/src/assembleHeadMat.cpp:            ThreadException e;
OpenMEEG/src/operators.cpp:        ThreadException e;
OpenMEEG/src/operators.cpp:        ThreadException e;
OpenMEEG/src/operators.cpp:        ThreadException e;
```

papadop:
OK, I have just pushed a solution to the memory management problem.
For the record:
- I used the PyCapsule idea documented in the numpy memory management documentation, but there were some subtleties linked to the use of C++.
- I had to add a method in Matrix to get an access to the shared_ptr containing the values. This method only exists for swig (i.e. protected by SWIGPYTHON guards).
- We will probably discover similar problems if we use the flat_array method for symmatrices, and a similar solution has to be used.
- The use of shared_ptr makes things a little bit more contrived because there no direct way of manipulating the reference counter (as we add with our own classes before moving to shared_ptr). See the line with the new in the additions.
- I also noticed that we were not consistent with the minimum numpy version. We asked for 1.14 in our code but the inclusion of numpy.i was putting it back to 1.7. I do not know how to solve that without modifying numpy.i, so I took that solution. A small advantage is that we now avoid including numpy/arrayobject.h twice (actually it was the use of SWIG_FILE_WITH_INIT which was bad.

Let's see what the builders say. I works on fedora 36.

wrapping/python/openmeeg/numpy.i
```diff
@@ -40,7 +40,10 @@
 #define NO_IMPORT_ARRAY
 #endif
 #include "stdio.h"
-#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
```
larsoner, line 43: This file is vendored directly from https://github.com/numpy/numpy/blob/main/tools/swig/numpy.i . Can we avoid modifying it?

larsoner:
> The use of shared_ptr makes things a little bit more contrived because there no direct way of manipulating the reference counter (as we add with our own classes before moving to shared_ptr). See the line with the new in the additions.

Could we always copy when coming from Python, and always copy when returning? This was the direction @agramfort and I went with `array_flat`. It avoids any sharing issues. It's obviously less efficient, but given the difficulties with getting the shared pointers right, it seems like it might make sense to do it. I don't foresee any huge performance hits or memory explosions from doing this, which would be the potential risk.

wrapping/python/openmeeg/numpy.i
papadop: I know, but I do not see how. The only possibility is to forget about requiring the 1.14 numpy API and stick with the 1.7 one. I have no idea who, why and when the 1.14 requirement was added.... The only possibility I can see would be to keep the diff as a patch and apply it, but I'm not sure this is worth for such a small change (we could even remove numpy.i from the repo and always download the latest version before applying the patch).

papadop:
> Could we always copy when coming from Python, and always copy when returning? This was the direction @agramfort and I went with `array_flat`. It avoids any sharing issues. It's obviously less efficient, but given the difficulties with getting the shared pointers right, it seems like it might make sense to do it. I don't foresee any huge performance hits or memory explosions from doing this, which would be the potential risk.

I do not think so. As I said before, the geom structure needs some internal consistency because point indexes are inferred by the distance to the beginning of the array. This means that the manipulated vertices must come from the original array, it cannot be a copy. Furthermore, we have already had quite a few reports where people were exploding their computer memory with complicated models. I do not really want to offer a second hand service to python users where heavy models can run in pure C++ and fail in python. This is the recipe for many complaints. The shared_ptr solution is not that bad, once it is found. Finally, if we copy, then we will probably raise another can of worms, because we will have incompatible python and OM structures. We already do that for small data structures (vertices, vectors, ...) and I do not think this is the right thing to do. We need the python to reflect the exact OM data, otherwise a mix of python changes and OM calls is doomed.

By the way, I'm not sure why you introduced array_flat (I thought I/we did that), given that this is of little use in python IMHO.

papadop:
Concerning the failures on windows and Mac, they are strange because they seem to point at the fact that there is still a copy of the array data.... Since I had some troubles with the swig modifications not always been taken into account, and given that I have seen that you have added some caching to the builds, could it be that there the caching interferes somehow ??

I can try to put messages to check whether the new openmeeg.i has been taken into account, but this is boring. Maybe I will fire my Windows virtual machine to look at that....

papadop:
A final comment is that test_python is awfully slow (546s out of 1004 for all tests). Can we make at least this test optional ?

wrapping/python/openmeeg/numpy.i
larsoner: > The only possibility is to forget about requiring the 1.14 numpy API and stick with the 1.7 one

This gets my vote, unless we need 1.14 capabilities for some reason

larsoner:
> we have already had quite a few reports where people were exploding their computer memory with complicated models

Is this due to the Mesh/Geometry spec inputs, or due to internal computations that happen after that? If it's the latter, then a quick copy before doing any heavy lifting should not be problematic.

> Finally, if we copy, then we will probably raise another can of worms, because we will have incompatible python and OM structures. We already do that for small data structures (vertices, vectors, ...) and I do not think this is the right thing to do. We need the python to reflect the exact OM data, otherwise a mix of python changes and OM calls is doomed.

If I create an OpenMEEG structure, then change (in-place) the `ndarray` they passed -- say the vertex locations, and/or the triangle definitions -- is OpenMEEG going to work properly in all cases? If not, then we're just in a different dangerous situation by trying to share pointers. Saying "we always make and use an internal copy of any data passed" seems safer than saying "we might store a ref to any ndarray passed, so don't change it or things might break in silent or explosive ways". That being said, if we can get the shared pointer stuff to work, then I can live with it.

> By the way, I'm not sure why you introduced array_flat (I thought I/we did that), given that this is of little use in python IMHO.

We need a way to persist just the BEM solution values to disk in FIF format, i.e., we need access to the 1D array after the computations are complete. Then we need to be able to reconstruct it later. If there was already a way to get these values in an `ndarray` then indeed we don't need `array_flat`. You can see it being used in `test_symmatrix.py`, so feel free to replace the uses there.

> Concerning the failures on windows and Mac, they are strange because they seem to point at the fact that there is still a copy of the array data.... Since I had some troubles with the swig modifications not always been taken into account, and given that I have seen that you have added some caching to the builds, could it be that there the caching interferes somehow ??

Only two caching mechanisms are in place

1. `ccache` -- if there is a problem here, then it would be very surprising since it's very widely used (and I don't think it works on Windows at the moment anyway!)
2. Caching of the `vcpkg` builds -- this is just for hdf5/matio/zlib, so shouldn't affect the builds

So I don't think it should be a caching problem.

> A final comment is that test_python is awfully slow (546s out of 1004 for all tests). Can we make at least this test optional ?

Sure, or maybe better is to replicate the error using the existing test data instead of adding a new test dataset. Presumably the `load_from_numpy` could be implemented equivalently there, and if it fails there as well, we can get rid of the sample stuff.

larsoner:
> Sure, or maybe better is to replicate the error using the existing test data instead of adding a new test dataset. Presumably the load_from_numpy could be implemented equivalently there, and if it fails there as well, we can get rid of the sample stuff.

Actually it looks like all tests already show both failures:
```
FAILED ..\..\Miniconda\envs\test\lib\site-packages\openmeeg\tests\test_python.py::test_python[True-Head1]
858
FAILED ..\..\Miniconda\envs\test\lib\site-packages\openmeeg\tests\test_python.py::test_python[True-mne_sample_ico3]
```
So I pushed a commit to skip the sample test. We can always reenable it later if we want

larsoner:
... and I pushed a workaround that might make things work (?): checking for values via `vertex(i)==V` instead of memory address via `&vertex(i)==&V`. It's not elegant and going to be slower (by a factor of 3?), but assuming there are no duplicate-location vertices in the mesh it might at least allow us to get something working in `main` while we pursue the more efficient correct solution.

larsoner:
Okay that did not work. To me it suggests that the actual locations are not recovered/stored properly. Which could still be some sort of ref counting problem, not sure. :( I'll revert 9cfdd517509607c588db5533c981eb9816b24c7d, then @papadop feel free to continue...

larsoner:
I'm starting to think this isn't about copy vs non-copy. I changed it to use the code path that just reads x/y/z `double` directly from the `vertices` ndarray, and then does `indmap.insert({ i, geom->add_vertex(Vertex(x,y,z)) });`, so a copy is already being made. But I'm not totally sure...

larsoner:
Okay I tried a bunch of random stuff and am still stumped :(

larsoner:
Okay it looks like the traceback is essentially:
```
C:\Miniconda\envs\test\lib\site-packages\openmeeg\tests\test_python.py:174: in test_python
502
    hm = om.HeadMat(geom, integrator, True)
HeadMat instantiation:
Assembling Head Matrix
    Assembling Cortex x Cortex (self) setting blocks
OPERATOR S ... (arg : mesh Cortex , mesh Cortex )
OPERATOR N ... (arg : mesh Cortex , mesh Cortex )
 done
    Assembling Skull x Cortex setting blocks
OPERATOR S ... (arg : mesh Skull , mesh Cortex )
OPERATOR N ... (arg : mesh Skull , mesh Cortex )
E   RuntimeError: Unknown vertex in triangle: -0.483673 0 0.782599
```
So for some reason it dies during the N operator. Why it would do this during N and not S?

larsoner:
Just to try to rule out some parallelization issues, I commented out all OpenMP stuff, but we should reenable it at some point. I also tried to make the message more verbose, now we get an access violation :(

larsoner:
Okay well it at least spits out some information now:
```
E   RuntimeError: Unknown vertex in triangle: Outer loop mesh Cortex requested vertex address:
E     0000014885BA3440 (-0.457386 0 0.740066)
E   but valid triangle vertex addresses are:
E     0000014885BA3980
E     0000014885BA3B00
E     0000014885BA3B40
```
but the fact that it gives an access violation when I ` << V` those candidate vertices seems to suggest that at least one of those addresses does not point to a valid Vertex.

wrapping/python/openmeeg/openmeeg.i
```diff
@@ -541,17 +575,21 @@ namespace OpenMEEG {
...
                 throw Error(SWIG_TypeError, "Geometry constructor list of lists must each have first entry a non-empty string.");
-            Mesh& mesh = geometry->add_mesh();
+            Mesh& mesh = geometry->add_mesh(PyUnicode_AsUTF8(name));
             PyObject* triangles = PyList_GetItem(item,2);
             mesh_add_triangles(&mesh,triangles,indmap[i]);
             mesh.update(true);
+            std::ostringstream oss;
+            oss << "SWIG Geometry update of meshes(" << i << ")=\"" << mesh.name() << "\"";
+            for (const auto& mesh : geometry->meshes())
+                mesh.check_consistency(oss.str());
```
larsoner, line 590: Okay finally some progress I think -- there are now *two* failures using the `check_consistency` function, which is hopefully temporary but just iterates over all edges to try to recreate the indexing problem. It does so, as expected only on Windows, but it seems to happen earlier than I expected, here during creation of the *second* mesh (Skull) it messes up the Cortex mesh, which was previously consistent (well before even computing S or N):

```
Create Geometry from list of lists
    geom_add_vertices(geom, pyobj) done.
    geom_add_vertices(geom, pyobj) done.
    geom_add_vertices(geom, pyobj) done.
    mesh_add_triangles for mesh Cortex : reference_vertices : triangles : check done.
Mesh Cortex (00000143F55713D0) consistent during SWIG Geometry update of meshes(0)="Cortex"
    mesh_add_triangles for mesh Skull : reference_vertices : triangles : check done.

E   RuntimeError: Unknown vertex in triangle: Mesh Cortex invalid    during SWIG Geometry update of meshes(1)="Skull", requested triangle vertex address:
E     00000143F5589900 (00000143F7A56420)
E   but valid triangle vertex addresses are:
E     00000143F7A56960
E     00000143F7A56AE0
E     00000143F7A56B20
```
or even worse/weirder in the other failure:
```
E   RuntimeError: Unknown vertex in triangle: Mesh Cortex invalid    during SWIG Geometry update of meshes(1)="Skull", requested triangle vertex address:
E     00000143F4B242A0 (00000143F557CEC0)
E   but valid triangle vertex addresses are:
E     00000143F557D5C0
E     0000000000000000
E     0000000000000000
```
https://github.com/openmeeg/openmeeg/runs/8181109365?check_suite_focus=true

I'm going to stop now, @papadop feel free to take over if you have some ideas here!

larsoner:
> but not on windows as I have not been able to create a proper lapack/blas environment

Ahh there is a script for this, too. I'll add it to https://github.com/openmeeg/openmeeg/pull/587, and try following the instructions all the way through locally to make sure they work.

> The problem was on the tag reading of files (used to determine the type of file)

That's weird since the Windows failures are on the code path where we read the files in Python and then pass list-of-lists with `ndarray` directly to `Geometry`. I guess we'll see...

larsoner:
Okay so now we have the two failures from before, plus a new one :)

papadop:
OK, not sufficient, I made my tests before merging and on test_python.
I'll try the sanitize technique on test_make_geometry_head...

larsoner:
@papadop I'll take care of the `merge` then push

larsoner:
Okay merge looks good, failures seem real/related

https://github.com/openmeeg/openmeeg/runs/8191980344?check_suite_focus=true#step:20:518

papadop:
Trying to follow things:
This code:
```
# First add mesh points

    indmaps = dict()
    geom = Geometry()
    for name, mesh in meshes.items():
        indmaps[name] = geom.add_vertices(mesh[0])

    # Create meshes

    for name, mesh in meshes.items():
        om_mesh = geom.add_mesh(name)
        om_mesh.add_triangles(mesh[1], indmaps[name])
        om_mesh.update(True)

```
was changed to:
```
# First add mesh points
    for name, mesh in meshes.items():
        if isinstance(mesh, Mesh):
            meshes[name] = _mesh_vertices_and_triangles(mesh)
        elif isinstance(mesh, (list, tuple)):
            pass
        else:
            raise ValueError(
                f"Wrong argument (should be a Mesh or a tuple of "
                f"vertices and triangles). Got {type(mesh)}"
            )

    # Create geometry from list-of-lists
    meshes = [[name, mesh[0], mesh[1]] for name, mesh in meshes.items()]
    geom = Geometry(meshes)
    del meshes

```

with
```
def _mesh_vertices_and_triangles(mesh):
    mesh_vertices = mesh.geometry().vertices()
    vertices = np.array([vertex.array() for vertex in mesh_vertices], np.float64)
    mesh_triangles = mesh.triangles()
    triangles = np.array(
        [mesh.triangle(triangle).array() for triangle in mesh_triangles],
        dtype=np.int64,
    )
    return vertices, triangles
```
I'm not sure I follow everything.... but one thing that strikes me is the absence of indmap.
This is needed because some meshes may share points, so we need to add the points of all meshes before adding triangles, because some triangles may require rewriting some indices (typically the shared ones, which will have a single representation in the vertices). The C++ code works in the same way, and this is why I added the indmap stuff and why meshes cannot be added fully one by one. We first need to add all the vertices, and then once this is set, we can start adding triangles eventually renaming some vertices...

That being said, for nested meshes, which are surely what is tested, this should not be a problem. On the other hand, the _mesh_vertices_and_triangles also makes copies of the arrays which should not work even on linux or MacOS. Can you explain why you removed this indmap stuff  ? Can we revert these changes ?

papadop:
To continue (and probably, I understand better why you did these changes):
test_make_geometry_head cannot work this way. It is not possible to read meshes independently, and then to aggregate them like that in a geometry. A mesh, can actually exist only in a Geometry.  When meshes are created like you did independently, each mesh has its own Geometry, which is a recipe for failures.

I did the interface with the idea that mne directly wants to provide its own meshes in terms of vertices and triangles. I did not suspect that you might want to use the om.Mesh to read files. If this is just for testing, then we can use om.Mesh but need to create numpy arrays from those to insert them in a geometry properly (i.e. not as meshes but as numpy lists of vertices and triangles.

papadop:
Yeah add_mesh might better be protected. It is dangerous to use if you do not understand well how meshes are working....
See comment at the beginning of the method:
`/// Add a mesh (assumes geometry points are of sufficient size).
`

wrapping/python/openmeeg/tests/test_python.py
```diff
@@ -19,39 +19,147 @@
...
 
-    geom = om.Geometry(geom_file, cond_file)
-    assert geom.is_nested()
+    if load_from_numpy:
+        mesh_files = read_geom(geom_file)
+        meshes = []
+        for fname in mesh_files:  # we assume the order is form inner to outer
+            points, _, tris = read_tri(fname)
+            meshes.append((points, tris))
+        geom = om.make_nested_geometry(meshes, conductivity=(1, 0.0125, 1))
```
larsoner, line 136: Notice that what gets passed in the `load_from_numpy` case is `geom` created by `make_nested_geometry`, which takes a list of tuple of ndarray `(points, tris)`, *not* OpenMEEG `Mesh` objects.

wrapping/python/openmeeg/_make_geometry.py
```diff
@@ -63,17 +64,10 @@ def make_geometry(meshes, interfaces, domains):
...
-
-    # Create meshes
-
-    for name, mesh in meshes.items():
-        om_mesh = geom.add_mesh(name)
-        om_mesh.add_triangles(mesh[1], indmaps[name])
-        om_mesh.update(True)
+    # Create geometry from list-of-lists
+    meshes = [[name, mesh[0], mesh[1]] for name, mesh in meshes.items()]
+    geom = Geometry(meshes)
```
larsoner, line 69: ... and then it gets used directly to instantiate the `geom` from list-of-list-of-ndarray...

wrapping/python/openmeeg/openmeeg.i
```diff
@@ -527,6 +560,7 @@ namespace OpenMEEG {
 %extend OpenMEEG::Geometry {
 
     Geometry(PyObject* pylist) {
+        std::cout << "Create Geometry from list of lists" << std::endl;
```
larsoner, line 563: ... which means you end up in this code path, not any that deal with any OpenMEEG `Mesh` at the Python level. 

> def _mesh_vertices_and_triangles(mesh):

Let's just get rid of this if it's unsafe. The failing code path (or at least one of them) uses the list-of-list-of-ndarray approach, not this code, anyway. I'll push a commit to make it clearer that this is not used.

So one of the failures is exactly with the code path you suggest, I think:

> I did the interface with the idea that mne directly wants to provide its own meshes in terms of vertices and triangles.
>
> ... but one thing that strikes me is the absence of indmap.

In this code path, `indmap` is created in C++ in the loop below. If you look at the traceback of the failure, you can even see that it's in this code path, because it reflects the messages printed below, stitched back together in the order in which they actually occurred:

```
C:\Miniconda\envs\test\lib\site-packages\openmeeg\tests\test_python.py:136: in test_python
    geom = om.make_nested_geometry(meshes, conductivity=(1, 0.0125, 1))
C:\Miniconda\envs\test\lib\site-packages\openmeeg\_make_geometry.py:203: in make_nested_geometry
    geom = make_geometry(meshes, interfaces, domains)
C:\Miniconda\envs\test\lib\site-packages\openmeeg\_make_geometry.py:69: in make_geometry
    geom = Geometry(meshes)
C:\Miniconda\envs\test\lib\site-packages\openmeeg\openmeeg.py:1224: in __init__
    _openmeeg.Geometry_swiginit(self, _openmeeg.new_Geometry(*args))

Create Geometry from list of lists
    geom_add_vertices(geom, pyobj) done.
    geom_add_vertices(geom, pyobj) done.
    geom_add_vertices(geom, pyobj) done.
    mesh_add_triangles for mesh Cortex : reference_vertices : triangles : check done.
Mesh Cortex (000001FDCBC6F070) consistent during SWIG Geometry update of meshes(0)="Cortex"
    mesh_add_triangles for mesh Skull : reference_vertices : triangles : check done.

E   RuntimeError: Unknown vertex in triangle: Mesh Cortex invalid    during SWIG Geometry update of meshes(1)="Skull", requested triangle vertex address:
E     000001FDCE2F8B80 (000001FDCE1D1060)
E   but valid triangle vertex addresses are:
E     000001FDCE1D1A60
E     000001FDCE1D1880
E     000001FDCE1D19C0
```
So don't look at the `test_make_geometry_head` failures, but rather the `test_python[True-Head1]` failures, like:

https://github.com/openmeeg/openmeeg/runs/8191980344?check_suite_focus=true#step:20:541

In the first iteration where Cortex is added, it's fine. But then once Skull is added, Cortex is broken.

larsoner (commented):
@papadop I think you might be onto something with the `_mesh_vertices_and_triangles` stuff, but I don't think it can explain the `test_python[True-Head1]` failures, which is the critical use case. That code shouldn't ever hit `_mesh_vertices_and_triangles` as it should pass list-of-list-of-ndarray to `om.Geometry` directly. Can you confirm?

papadop:
> @papadop I think you might be onto something with the `_mesh_vertices_and_triangles` stuff, but I don't think it can explain the `test_python[True-Head1]` failures, which is the critical use case. That code shouldn't ever hit `_mesh_vertices_and_triangles` as it should pass list-of-list-of-ndarray to `om.Geometry` directly. Can you confirm?

I can confirm. Digging further, I found that this code just extracts a copy of vertices and triangles from om.Mesh. This should be OK.

wrapping/python/openmeeg/tests/test_python.py
papadop: OK.

larsoner:
Okay I had a thought about what might be happening. In `geometry.h` for `add_mesh(std::string name)` there is this comment:

> //  It is dangerous to store the returned mesh because the vector can be reallocated.
> //  Use mesh(name) after all meshes have been added....

I've tried splitting the loop into two parts that do exactly this. My guess/hunch is that the vector *is* being reallocated on Windows, but not macOS or Linux (for whatever reason), thus messing up the previous mesh... we'll see!

larsoner:
@papadop it looks like the macOS failure is related to your file code:
```
 /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibw-run-dll8425n/cp37-macosx_x86_64/venv-test/lib/python3.7/site-packages/openmeeg/tests/test_python.py:148: in test_python
      dipoles.load(dipole_file)
  E   OSError: Exception: Bad storage type in file /Users/runner/work/openmeeg/openmeeg/data/Head1/Head1.dip.
```
can you look?

In the meantime, splitting the loop seems to have shifted the error, suggesting (I think) we're on to something:
```
      geom.finalize()
  E   RuntimeError: Unknown vertex in triangle: Mesh Cortex invalid    during finalize() start, requested triangle vertex address:
  E     000001B85E6835A0
  E   but valid triangle vertex addresses are:
  E     000001B85E6844E0
  E     000001B85E684300
  E     000001B85E684440
```
I'm going to try a `mesh.clear()`, and a bit of reorg (create all meshes first, clear them, add verts, add triangles).

larsoner:
... actually `geometry->meshes().reserve(N);` is what I see in the I/O code, hopefully it's enough here.

larsoner:
Okay all failures are now of the form:
```
OSError: Exception: Bad storage type in file
```
@papadop once you fix this I think we'll be good to go!

larsoner:
For now I've done `git revert 43761ec77bf79f6ea166c7da289756bfd4581101` since that seems to be what broke file reading in the tests on Windows and macOS, where they all said before reverting:
```
E   OSError: Exception: Bad storage type in file TrivialBinIO linop.storageType (0) != inforead.storageType (4) for file C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\test_pointer_clearing0\Head1.dip.
```
@papadop feel free to reapply the commit and fix it if you want, but I'm inclined to tackle that in a separate PR if possible. For now the tests have been fixed by the revert I think.

The only other thing that's red is this on macOS:
```
    Start 363: cmp-EEGadjointEST-dip-Head1-dip3-mag
1/2 Test #363: cmp-EEGadjointEST-dip-Head1-dip3-mag ...***Failed    0.02 sec
-------------------------------------------------
Comparing : 
- /Users/runner/work/openmeeg/openmeeg/build/tests/Head1-dip.est_eegadjoint
- /Users/runner/work/openmeeg/openmeeg/tests/analytic/eeg_head1_analytic.txt
ERROR MAG (column 2 ) = 0.803454	relMAG = 0.196546

ERROR : Matrices are different at the precision : eps = 0.13

    Start 366: cmp-EEGadjointEST-dip-Head1-dip3-rdm
2/2 Test #366: cmp-EEGadjointEST-dip-Head1-dip3-rdm ...***Failed    0.02 sec
-------------------------------------------------
Comparing : 
- /Users/runner/work/openmeeg/openmeeg/build/tests/Head1-dip.est_eegadjoint
- /Users/runner/work/openmeeg/openmeeg/tests/analytic/eeg_head1_analytic.txt
ERROR RDM (column 2 ) 0.46107522558215308095

ERROR : Matrices are different at the precision : eps = 0.13
```
Once we get that green we should be able to merge. But I'm not totally sure what's required to fix it.

larsoner:
Okay:

1. The "big problem" was not preallocating the meshes, or `add_mesh`ing then modifying their geometry in the same loop, which would sometimes cause reallocation of the vector that stored them, thereby breaking the meshes. This has been solved most cleanly by a new `Geometry(nMeshes)` constructor, where you preallocate the space for the meshes vector. 
2. The macOS failures were due to OpenMP. With it enabled, an EEG test fails. With it disabled, it passed. Thus I've disabled it for now, like we do for the wheels. I think we should investigate this separately.
3. I restored 43761ec77bf79f6ea166c7da289756bfd4581101 then rolled a tiny bit of it back. This was enough to get things green.

@papadop let me know if you're happy. The more progress we can make in follow-up PRs, the better.

We should squash-and-merge by removing all intermediate commit names, but keeping the `co-authored by` lines for me and @papadop since this was a real team effort to figure out!

agramfort (commented):
🙏 @larsoner and @papadop for the team effort !

@papadop I let you squash and merge (please clean a bit the log of the commit as suggested by @larsoner)

🙏 🙏 🙏

papadop:
> 3\. I restored [43761ec](https://github.com/openmeeg/openmeeg/commit/43761ec77bf79f6ea166c7da289756bfd4581101) then rolled a tiny bit of it back. This was enough to get things green.
> 
> @papadop let me know if you're happy. The more progress we can make in follow-up PRs, the better.

Actually most of the commit is kept... Agree to look at that in a follow-up PR.

papadop:
Will push my changes and wait for the checkers before merging.

papadop:
Oh yeah, I also gave an unsigned parameter to the new Geometry constructor and thus removed the exception.

papadop:
A final question I have is that you have added explicit array types in python in several places. I suppose/hope that there are no differences across platforms in python types... You know better and have more experience than I do, with this respect. I just wanted to mention the thing.

agramfort:
@papadop we should be good and if not the CIs should tell us. I let you push your minor changes and merge when CIs are green with tiny edits to commit message 🙏

papadop:
Oh yeah, another small change is the naming of OpenBLAS. It was OpenBlas in the cmake option but we were testing for OpenBLAS !!! Seems not to have been a problem, but for consistency I changed it. All the builders, are using OpenBLAS already, so this should not interfere....

papadop:
OK, pushed my last changes. I hope the testers will gobble them without shouting !!!!
Then I will squash and merge.

papadop:
Thank's a lot Eric. It was indeed a true collaborative work !!!

larsoner:
Sounds good, I marked for auto squash and merge when green!

papadop:
I forgot to say that your change @larsoner in make_geometry using the Geometry constructor was fine in principle. Feel free to re-submit it in another PR. It has the advantage that we would not need to expose the internal functions add_vertices and add_geometry to python (even though I'm unsure on how to do that). If it is equivalent memory consumption wise and elapsed time wise, we can re-introduce it.

agramfort:
thanks @larsoner and @papadop !

PR #589, opened 2022-09-02
ENH: Add verbose argument to HeadMat, expose Integrator

larsoner:
When running in Python, these outputs so far are mandatory:

https://github.com/mne-tools/mne-python/runs/8145144458?check_suite_focus=true#step:13:602

```
mne/forward/tests/test_make_forward.py::test_make_forward_solution_ctf[testing_data-MNE-C] PASSED [ 15%]
mne/forward/tests/test_make_forward.py::test_make_forward_solution_ctf[testing_data-openmeeg] OPERATOR S ... (arg : mesh Cortex , mesh Cortex )
OPERATOR N ... (arg : mesh Cortex , mesh Cortex )
OPERATOR S ... (arg : mesh Skull , mesh Cortex )
OPERATOR N ... (arg : mesh Skull , mesh Cortex )
OPERATOR D ... (arg : mesh Skull , mesh Cortex )
OPERATOR D*... (arg : mesh Skull , mesh Cortex )
OPERATOR S ... (arg : mesh Skull , mesh Skull )
OPERATOR N ... (arg : mesh Skull , mesh Skull )
OPERATOR D ... (arg : mesh Skull , mesh Skull )
OPERATOR S ... (arg : mesh Head , mesh Skull )
OPERATOR N ... (arg : mesh Head , mesh Skull )
OPERATOR D*... (arg : mesh Head , mesh Skull )
OPERATOR S ... (arg : mesh Head , mesh Head )
OPERATOR N ... (arg : mesh Head , mesh Head )
PASSED [ 15%]
mne/forward/tests/test_make_forward.py::test_make_forward_solution_discrete[testing_data] PASSED [ 15%]
```

This adds a `verbose` argument in the right places so that (I think!) you should be able to do `HeadMat(geo, integrator, False)` in Python to avoid these messages.

It also ~~adds a note about why Integrator can't be exposed. I'll make a clearer note about this in the "python issues" issue.~~ now exposes `om.Integrator` so that it can be passed in.

### DIFF
--- a/wrapping/python/openmeeg/tests/test_python.py
+++ b/wrapping/python/openmeeg/tests/test_python.py
@@ -53,11 +53,11 @@ def test_python(data_path, tmp_path):
     n_eeg_sensors = patches.getNumberOfSensors()
 
     # Compute forward problem (Build Gain Matrices)
-    # gauss_order = 3  # XXX cannot get Integrator exposed
+    # gauss_order = 3  # XXX cannot get Integrator exposed because of SyntaxError when doing %include <integrator.h>
     # use_adaptive_integration = True
     # dipole_in_cortex = True
 
-    hm = om.HeadMat(geom)
+    hm = om.HeadMat(geom, False)
     hminv = hm.inverse()  # invert hm with a copy
     hminv_inplace = om.HeadMat(geom)
     hminv_inplace.invert()  # invert hm inplace (no copy)
--- a/wrapping/python/openmeeg/tests/test_python.py
+++ b/wrapping/python/openmeeg/tests/test_python.py
@@ -57,7 +57,7 @@ def test_python(data_path, tmp_path):
     # use_adaptive_integration = True
     # dipole_in_cortex = True
 
-    hm = om.HeadMat(geom, False)
+    hm = om.HeadMat(geom, om.Integrator(3, 0, 0.005), False)
     hminv = hm.inverse()  # invert hm with a copy
     hminv_inplace = om.HeadMat(geom)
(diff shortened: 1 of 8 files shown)


wrapping/python/openmeeg/tests/test_python.py
```diff
@@ -53,17 +53,18 @@ def test_python(data_path, tmp_path):
     n_eeg_sensors = patches.getNumberOfSensors()
 
     # Compute forward problem (Build Gain Matrices)
-    # gauss_order = 3  # XXX cannot get Integrator exposed
+    # gauss_order = 3  # XXX Integrator is now exposed, just need to use it...
```
larsoner, line 56: @agramfort, in case you want to take a stab at this, it should be available after this PR is in

agramfort (approved):

wrapping/python/openmeeg/tests/test_python.py
papadop: I like the idea of adding a control over verbosity. I like less the need to pass it as a constructor parameter.
Could we instead add a set_verbose method that allows to change the verbosity at will.
We could even try to make different default for different modes (off for python and on for C++ apps to keep backward compatibility)....

Any opinion ?

larsoner: In part the motivation behind having it in the constructor was because it was already this way elsewhere in the codebase

At least for HeadMat, it has to be in the constructor because all the computations are done immediately when you call the constructor

agramfort:
+1 to have it in constructor. It's much easier to discover and our constructors are not too crowded.

+1 to MRG as is

PR #591, opened 2022-09-07
add debug files and script comparing to nolte's method from fieldtrip

agramfort:
work in progress...

the script gives me for corrs descriptive stats:

```
>>>print(np.min(corrs), np.max(corrs), np.mean(corrs), np.median(corrs))
-0.6002644289808833 0.9995001452346635 0.969677137485964 0.9938644798114129
```

So numbers are pretty good but some are really bad.

Let's see what the MNE solver says... but later

it's just FYI @papadop @larsoner no need to react yet


agramfort:
I do confirm that Fieldtrip is a lot more similar to MNE than OpenMEEG when using a single layer model. See the figure of correlations:

hist_corrs_mne_ft_om

agramfort:
and here is the result when I take an openmeeg 3 layers model vs 1 layer in FT or MNE.

hist_corrs_mne_ft_om3

So you see that using an improper head model improves the correlations which I cannot explain...

@papadop does it help?

papadop:
I will try to make a similar test using the C++ commands. 
I want just to rule out a possible python mistake somewhere even though that does not seem very probable....

agramfort:
Let me know if you want the files produced by fieldtrip and mne

Issue #584, opened 2022-08-30
BUG: Bugs with Python wrappers


larsoner:
- [ ] Our Python API is not consistent. Sometimes you accept an array sometimes you don’t. If you screw up the shape it segfaults 
- [ ] If you don’t have a Fortran layout it should raise a nice exception
- [ ] We need to test better single layer model (#577)
- [ ] Constructing a Sensors object in Python is a mess
- [ ] Fix XXX in the tests

_Adapted from code posted by @agramfort in https://github.com/openmeeg/openmeeg/issues/564#issuecomment-1231541158_

papadop:
For constructing sensors, we have to adopt the Geometry strategy (see %extend OpenMEEG::Geometry at the end of openmeeg.i). I think this is the only way to have a nice pythonic style interface....

agramfort:
@papadop I find the constructors for Sensors class quiet confusing now. I am plead guilty to some extend. Basically we support labels or not, weights or not etc.

larsoner:
On Windows it seems like when doing `om.Geometry(geom_file, cond_file)`, the two files cannot be deleted afterward:

https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=21617&view=logs&j=305851a9-a7bb-55db-0042-7e2b6f48aa1c&t=4d37777d-f36a-53aa-9217-6386d15dddcd&l=3479

```
E   PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\openmeeg-io-t2wktag7\\tmp.geom'
```

These should be read immediately on `Geometry` instantiation then the file pointers closed, right? It seems like this is not the case for Windows.

larsoner:
In an effort to expose `om.Integrator` in Python, I tried adding this to `openmeeg.i`:
```
%include <integrator.h>
```
but SWIG is not happy about it:
```
/Users/larsoner/python/openmeeg/OpenMEEG/include/integrator.h:45: Error: Syntax error in input(3).
```
which points to this line

https://github.com/openmeeg/openmeeg/blob/b00b2dc295a7f8923600f2f7adb06554ba83e26d/OpenMEEG/include/integrator.h#L45

[...]

PR #592, opened 2022-09-07
MAINT: Clean up imports

larsoner:
# Idea

Improve maintainability by reducing the number of public-level C++ classes available.

# Implemented so far

1. Rename the SWIG-generated `openmeeg/openmeeg.py` to a private `openmeeg/_openmeeg_cxx.py`
2. Replace `from .openmeeg import *` in `__init__.py` with `from ._openmeeg_cxx import ...` for as few things as possible
3. Remove stuff from the namespace like `Vector` and `Mesh` that users shouldn't need to use (they should work with `ndarray` directly)
4. Replace `om.Geometry(geom_file, cond_file)` with `om.read_geometry(geom_file, cond_file)` so that we can do validation in Python.

Point (4) critically allows us to and avoid exposing the `Geometry` class constructors in Python, as users should now either `read_geometry`, `make_geometry`, or `make_nested_geometry`, all of which we control at the Python level and make sure `Geometry` gets constructed correctly.

~~This PR is WIP because I know there are some refs in CIs that will need to be updated to this new scheme, so I'll need a few iterations to clean those up.~~

# Proposal for ~~this or~~ follow-up PRs

Ideally I think we would have *no* `from ._openmeeg_cxx import ...`s in `__init__.py` / the public interface at all, and instead go the route taken here with `Geometry`. For `om.Sensors`, for example, it would be nice to have just a `read_sensors` that works with filenames and a `make_sensors` that works with `ndarray`, without a bunch of extra gymnastics to satisfy the C++ classes. This approach would allow us to tackle these three limitations from #584 at once with relative ease:

> - Our Python API is not consistent. Sometimes you accept an array sometimes you don’t. If you screw up the shape it segfaults
> - If you don’t have a Fortran layout it should raise a nice exception
> - Constructing a Sensors object in Python is a mess (weights, radii, etc.)

Instead of modifying the C++ code to be smarter, in Python let's check all sizes and dtypes, convert to the Fortran layout the C++ code wants (telling people that for max memory efficiency they should construct their `ndarray` in this layout). To me this is much more sustainable than modifying the C++ code, and makes it easier for others to contribute fixes and enhancements.

@agramfort @papadop do you think this is worthwhile? If so I can implement this fairly quickly I think. I think it would also warrant a `3.0` release since it's a big backward compat break to remove all of these in the Python bindings.

FWIW, the BEM solution is pretty much already covered by `make(_nested)_geometry`, so we'd "just" need to convert these I think:

https://github.com/mne-tools/mne-python/blob/cf235d539eb932013ea1fa5b78bc2dbf259e1c96/mne/forward/_compute_forward.py#L834-L865

### DIFF
--- a/build_tools/cibw_before_build_windows.sh
+++ b/build_tools/cibw_before_build_windows.sh
@@ -26,5 +26,5 @@ python -m pip install "$NUMPY_PIP" --only-binary="numpy"
 cmake -B build -DENABLE_PYTHON=ON -DPython3_EXECUTABLE="$(which python)" .
 cmake --build build --config Release
 python -m pip uninstall -yq numpy
-cp -av build/wrapping/python/openmeeg/*.pyd build/wrapping/python/openmeeg/openmeeg.py wrapping/python/openmeeg/
+cp -av build/wrapping/python/openmeeg/*.pyd build/wrapping/python/openmeeg/_openmeeg_cxx.py wrapping/python/openmeeg/
 rm -Rf build
--- a/wrapping/python/setup.cfg
+++ b/wrapping/python/setup.cfg
@@ -1,6 +1,6 @@
 [flake8]
 ignore = W503
-exclude = openmeeg/openmeeg.py
+exclude = openmeeg/_openmeeg_cxx.py
 per-file-ignores =
     __init__.py:F401
     setup.py:E501


agramfort (commented):
I ❤️ the idea !

larsoner:
This one is green so if you're okay with moving in this direction @papadop feel free to review and merge and I can do the rest in a follow-up PR to keep things readable

papadop:
I like the goals and the main philosophy of this change.
I'm less sure on the way it is implemented. I need to review the whole set of changes to be sure.
A quick look makes me a little bit afraid as I'm seeing strange cp xxx yyy in the build scripts, I need to understand better.

I wonder, whether why some or most of this couldn't be implemented handled directly in swig using %ignore and %rename.
But I need to understand better the changes for this. In particular, I'm not sure to understand what the renaming of openmeeg.py into openmeeg_cxx.py buys us. We should just be able to choose what we import in __init__.py, (and eventually remove non useful stuff with %ignore or rename private stuff with %rename in swig).

I really need to have a close look to the change.

papadop:
Just as a notice....
I expect to push a WIP branch during the week-end, that starts to implement the apps in python with a git like syntax:
om assemble ...
om gain ....

I suspect that there will be some redundancy with the complete pipelines that you have built so far. We can probably share some functions....

larsoner:
> I'm less sure on the way it is implemented. I need to review the whole set of changes to be sure.

The list in the top comment should hopefully make it clear what this PR does so far. The only user-facing changes are to make `openmeeg.openmeeg` private to `openmeeg._openmeeg_cxx`, and then to not use `from .openmeeg import *` but instead `from ._openmeeg_cxx import ...` with a smaller explicit list. But eventually the idea is that *none* of these CXX classes will be exposed to users (i.e., not be in `openmeeg.__init__` and therefore not in the `openmeeg` namespace).

> I expect to push a WIP branch during the week-end, that starts to implement the apps in python with a git like syntax: om assemble ...

It might be better to hold off. I actually don't think we need this if we make more usable Python code like `read_geometry(..., ...)` or `make_sensors(..., ...)`. Command-line utils with file-based I/O is much less Pythonic than function-based stuff. So I'm not sure it adds much value compared to working on and supporting proper Python functions.

papadop:
> It might be better to hold off. I actually don't think we need this if we make more usable Python code like read_geometry(..., ...) or make_sensors(..., ...). Command-line utils with file-based I/O is much less Pythonic than function-based stuff. So I'm not sure it adds much value compared to working on and supporting proper Python functions.

The idea is (long term) to replace the C++ executables with python based commands. This is simpler to maintain and will exercise the python interface thoroughly. But as some of our partners are not yet ready to add a python dependency, it will just be some side stuff for now. Yet I have see that there are several ways to produce self contained .exe commands from python script, so it might even be a simpler way to package these commands. But I do not intend to go that far for now. 

> Actually from seeing @agramfort workflow recently and with #593 it seems like file I/O-based workflows are pretty common for people. I suppose we could add these as `om viz` / `om assemble` etc. Though it might be safer to use `openmeeg viz` and `openmeeg assemble`, then if people really want the `om` shorthand they can just `alias` it in the `.bashrc`.

I buy that argument... I will do that. I mean replace om by openmeeg.

papadop:
> The list in the top comment should hopefully make it clear what this PR does so far. The only user-facing changes are to make `openmeeg.openmeeg` private to `openmeeg._openmeeg_cxx`, and then to not use `from .openmeeg import *` but instead `from ._openmeeg_cxx import ...` with a smaller explicit list. But eventually the idea is that _none_ of these CXX classes will be exposed to users (i.e., not be in `openmeeg.__init__` and therefore not in the `openmeeg` namespace).

But what is the difference with keeping the module as it is now and importing only what we want  in __init__.py ?
I do not see the value of renaming the file... What does it buy us ?

larsoner:
> I do not see the value of renaming the file... What does it buy us ?

Anything not prefixed by an underscore in Python is considered public. So in `main` people can do this:
```
from openmeeg.openmeeg import Geometry
```
and their expectation is that this is supported. By renaming this SWIG C++ module to `openmeeg._openmeeg_cxx` we make everything in there private, rather than implicitly public.

agramfort:
And if it’s private API we can change it without the fear of breaking someone’s code and it will make clear in the mne code that we use public functions that openmeeg guarantees to not break without proper depreciation

larsoner:
@papadop okay to move forward here?

I could then clean up #593 and set up the infrastructure for `openmeeg ...` so it's easier for you to implement `openmeeg assemble ...` etc. command-line equivalents to `om_assemble` etc.

papadop:
OK. I have two choices here. Either we merge this change like this, but will change its content afterwards (see below), or we make the changes now and this will trigger some more delays.
About the changes, I would like to see:
- Instead of _openmeeg_cxx, rename instead the module openmeeg_private. I think this will make the message clearer.
- Make all the functions that are not need ignored at the swig level (with %ignore). This is a long term goal and is not relevant for the two choices I mentioned above, i.e. for this merge.
- As a prerequisite of the previous point (so not needed for this merge): Audit (make a list) of all currently exported functions and separate them in three classes: Those that should not be exported, those that should be private and those that should be public. This will allow us to define a clear API for the python module.

So the main question I have at this time: do you prefer we change the module name to openmeeg_private before merging or after it ?

papadop:
I actually have a provisional patch that does the _openmeeg_cxx -> openmeeg_private change. It was less work than I thought. So I may submit it, to see whether checkers are happy, and if for some reasons it is not appropriate, then we can revert it....

larsoner:
> _openmeeg_cxx -> openmeeg_private

Please do not remove the leading underscore in the name. In the Python community and to people who work with Python everyday / across many projects, the leading underscore is what says immediately "this is private", people don't really name modules `whatever_private`. I can live with `_openmeeg_private` if you insist on this pattern, but to me it's less standard and more redundant compared to `_openmeeg_cxx` (which already communicates the "private" status because of the community standards of leading underscore use)

agramfort:
+1 _ means private already. Let's just use _openmeeg_cxx

papadop:
OK. I tried to keep theinitial _ but without the private and this does not work. I must say that I do not understand at all the python mechanisms. I take the same file _openmeeg.py and copy it with the names openmeeg.py and _openmeeg_private.py (or _openmeeg_cxx.py). Then:
```
 >>> from _openmeeg import GainMEG
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'GainMEG' from '_openmeeg' (/home/papadop/compiles/OpenMEEG/OpenMEEG-ref-f36/wrapping/python/openmeeg/_openmeeg.so)
>>> from openmeeg import GainMEG
from _openmeeg_private import GainMEG
>>> 

```
But in all cases:
```
>>> from _openmeeg import HeadMat
>>> from openmeeg import HeadMat
>>> from _openmeeg_private import HeadMat
>>> 
```

So global vars are imported anyway, but not classes ?!?
In other words, why is the class GainMEG private for _openmeeg.py and not for _openmeeg_private.py, while the function HeadMat is public in both cases....

larsoner:
It depends on the folder structure and your current directory, which must have been `openmeeg`. Users will never do this in practice, though, so I wouldn't worry about it.

Ultimately I don't think this name matters much since it's not user facing (as long as it has the leading underscore), but CIs are now red. Can we just go back to `_openmeeg_cxx.py` and d249e8c so we can move forward with more user facing/ usability stuff in follow up PRs?

agramfort:
I reverted the commits that crash the CIs @papadop and @larsoner 
and merged so can proceed with making more things private.

papadop:
Thank's for having reverted for me. I'm still unconvinced by the way these goals are achieved (while agreeing with the goals). But let's advance.
I think I have a counter proposal which controls what is internal (as I learned that private is not a python concept) directly in the openmeeg module.

PR #593, opened 2022-09-09
add om_viz CL tool to view quickly some openmeeg head model files

agramfort:
@larsoner let me know if you see a better way to do this and test this.


larsoner:
I guess we could do this as `om viz` or (maybe better) `openmeeg viz` as I just wrote in https://github.com/openmeeg/openmeeg/pull/592#issuecomment-1242180184. So let's wait for the git-like syntax to work then we can add this.

We probably want to do some refactoring here as well with `read_mesh` / `read_tri` in a `openmeeg._utils` so that they aren't public (or we can choose to make them public if we want) and we can reuse them in `tests/test_python.py`. I'll push some commits for this probably on Monday

agramfort:
Just a warning that having "openmeeg foo" would be neat **but** it's not
really an option for brainstorm users
that only need the binaries. I would not require a python env for matlab
users until we are sure it's not a problem.
So the current apps for me must stay.

papadop:
Eric, this is python. Your call here. Personnally, I have nothing against adding this python script.... but I have not tested it in any way.

larsoner:
I would be inclined nowadays to use pyvistaqt + BackgroundPlotter. I can push a couple of commits

larsoner:
Okay that wasn't too bad. Locally:
```
larsoner:~/python/openmeeg/data/Head1$ om_viz --opacity 0.15 Head1.geom Head1.squids
```
Screenshot From 2026-07-28 14-04-24

I'll get CIs green then merge!

larsoner:
Oh the failure is just what https://github.com/openmeeg/openmeeg/pull/863 works around. Restarting the job

PR #595, opened 2022-09-20
WIP: Add first version of python applications.

papadop:
This is work in progress. Just to show the organisation and how it works.
I need at least to add some cmake stuff and to run tests.

I'm sure there are numerous parser small mistakes...

### DIFF
--- a/wrapping/python/apps/assemble.py
+++ b/wrapping/python/apps/assemble.py
@@ -1,0 +1,289 @@
+import sys
+import argparse
+#import openmeeg as om
+
+def parser(argv):
+    print(argv)
+    parser = argparse.ArgumentParser(prog=argv[0],description='Compute various leadfields (gain matrices).')
+    parser.add_argument('--old-ordering', type=bool, default=False, help='Using old ordering i.e using (V1, p1, V2, p2, V3) instead of (V1, V2, V3, p1, p2)')
+    subparsers = parser.add_subparsers(dest='subparser_name', help='sub-command help')
+
+    # create the parser for the HeadMat case
+    HMparser = subparsers.add_parser('HeadMat', aliases=['HM', 'hm'], help='Compute Head Matrix for Symmetric BEM (left-hand side of the linear system).')
+    HMparser.add_argument('geometry', type=open, nargs=1, help='head geometry file name (.geom)')
+    HMparser.add_argument('conductivities', type=open, nargs=1, help='head conductivities file name (.cond)')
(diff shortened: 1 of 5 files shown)


wrapping/python/apps/assemble.py
```diff
@@ -0,0 +1,289 @@
+import sys
+import argparse
+#import openmeeg as om
+
+def parser(argv):
```
larsoner, line 5: Rather than a single monolithic file I find MNE-Python's way of doing things much easier -- each `mne whatever` has a `mne/commands/mne_whatever.py` that you modify with your own arguments

https://github.com/mne-tools/mne-python/blob/main/mne/commands/mne_browse_raw.py

I think it's more maintainable than this approach, where everything lives in a single namespace

I can convert it to this format if you want

papadop: I'm not sure I understand. each `openmeeg command` already has its own `command.py` file. If you meanto make the same at each subcommand level i.e. `openmeeg command subcommand` into its own  `subcommand.py` file, then I totally agree. This will be particularly useful for the assemble and gain commands, which are huge.

What I just want to have (but I'm sure that is also done with mne) is to make sure that all:
```
openmeeg --help
openmeeg command --help
openmeeeg command subcommand --help
```
provide the relevant messages.

Note though that there are two commands in assemble that share the same `processing`, but that's largely doable with your proposed split. Also, I like better the run() name, so if we split at the subcommand level, there is no need to call each function with its own name, they can all have the same run method.

Did I understood your proposal correctly ?
If yes, be my guest, as I have other tasks to do for the next hours/days...
In particular, if you have some proper install example to follow from mne that would be great too (coming commit, but probably with wrong install places).

PS: I did not knew that mne-python had a command line interface !!! There is always something new to learn...

larsoner: > I'm not sure I understand. each openmeeg command already has its own command.py file.

Ahh I just misread / didn't look closely enough, I didn't realize there were `subcommand`s

> In particular, if you have some proper install example to follow from mne that would be great too (coming commit, but probably with wrong install places).

Usually with `setuptools` at least you provide an `entry_point`:

https://github.com/mne-tools/mne-python/blob/7730df72dc51f663fe557e997b6f5486111f3e43/setup.py#L138-L139

This should get picked up automatically when we build wheels, and it's easy to add to `conda` when we package there. I'm not sure how to do it with `cmake` actually, but hopefully it's not too hard there either...

PR #596, opened 2022-09-20
Python minor improvements

papadop:
Minor stuff:
- Some CMakeLists.txt small improvements.
- Make more symbols public in the python module (typically those that would be needed for python apps).

I hope that the testers will not choke on those (I would not understand why).

### DIFF
--- a/wrapping/python/openmeeg/__init__.py
+++ b/wrapping/python/openmeeg/__init__.py
@@ -2,25 +2,39 @@ from . import _distributor_init
 
 # Here we import as few things as possible to keep our API as limited as
 # possible
+
 from ._openmeeg_cxx import (
+    Matrix,
+    SymMatrix,
+    SparseMatrix,
+
     HeadMat,
     Sensors,
-    Integrator,
-    Head2EEGMat,
-    Head2MEGMat,
-    DipSourceMat,
+
+    CorticalMat,
+    CorticalMat2,
+    DipSource2InternalPotMat,
     DipSource2MEGMat,
+    DipSourceMat,
+    EITSourceMat,
+    Forward,
     GainEEG,
     GainMEG,
     GainEEGadjoint,
     GainMEGadjoint,
     GainEEGMEGadjoint,
-    Forward,
+    GainEITInternalPot,
+    GainInternalPot,
+    Head2ECoGMat,
+    Head2EEGMat,
+    Head2MEGMat,
     SurfSourceMat,
     SurfSource2MEGMat,
-    Matrix,
-    SymMatrix,
+    Surf2VolMat,
+
+    Integrator,
 )
+
 from ._version import __version__
 from ._make_geometry import make_geometry, make_nested_geometry, read_geometry
 from ._utils import get_log_level, set_log_level, use_log_level
--- a/wrapping/python/CMakeLists.txt
+++ b/wrapping/python/CMakeLists.txt
@@ -11,7 +11,6 @@
 # those macros are driven by CMAKE_SWIG_FLAGS, and SWIG_MODULE_XXXXXXX_EXTRA_DEPS
 # where XXXXXXX is the name of the target object in swig_add_library(XXXXXXX ...)
 
-
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development NumPy)
 if (BUILD_TESTING)
     find_package(Pytest)
@@ -28,7 +27,6 @@ include_directories(${Python3_INCLUDE_DIRS}
 )
 
 cmake_policy(SET CMP0078 NEW)
-set(SWIG_TARGET_NAME "openmeeg")
 cmake_policy(SET CMP0086 NEW) # Use NEW policy and explicitly use SWIG_MODULE_NAME
 
 find_package(SWIG REQUIRED)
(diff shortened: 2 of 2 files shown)


papadop:
Grrrr. black is a little bit pedantic. It forces me to remove white lines that were explicitly made to show groups of different type.

agramfort:
@papadop mac build is not happy.

wrapping/python/openmeeg/__init__.py
```diff
@@ -2,25 +2,36 @@
...
+    DipSourceMat,
+    EITSourceMat,
+    Forward,
     GainEEG,
     GainMEG,
     GainEEGadjoint,
     GainMEGadjoint,
     GainEEGMEGadjoint,
-    Forward,
+    GainEITInternalPot,
```
larsoner, line 24: See https://github.com/openmeeg/openmeeg/pull/594#issuecomment-1252684051, I'd rather this list eventually shrink to zero and instead provide Pythonic interfaces in the public API (using C++ APIs privately) as much as possible rather than expose, make public, and need to support indefinitely (or deal with deprecation cycles, etc.) any C++ classes to Python users. To me it makes maintenance and contributing both easier

papadop:
> @papadop mac build is not happy.

I think it's a network issue... I restarted the job.

wrapping/python/openmeeg/__init__.py
papadop: I'm not opposed to that. But for the time being, those functions/classes are needed. 

Note though that it is not clear to me what would bring the pythonic interfaces over the existing ones. Most of those are simple functions, or classes with very few methods. They are very different from Mesh, etc for which I totally agree.

Maybe you can show me an example of what you mean.

larsoner: > Most of those are simple functions, or classes with very few methods. They are very different from Mesh, etc for which I totally agree.

But why do end-users need them at all? They are designed to make things easier in C++ and are barely (or not at all?) validated for use with Python, *and* in Python things like Vect3 at least are pointless because of NumPy.

One root problem/difference is that people in Python are used to being loose with types (duck typing), and C++ is not this way. In python, you can pass a list of lists of floats and have it be interpreted properly as ndarray. You can pass non-contiguous arrays. You can often pass arrays of the wrong dtype, and things will be automatically converted/dealt with for you.

In C++ in general -- and OpenMEEG seems to be no exception -- you get any of these wrong, your code blows up, sometimes with a nice error, but often in some worse way (segfault) or a dangerous one (silently giving a wrong answer). This was @agramfort's experience trying to use the classes, which prompted the comments in https://github.com/openmeeg/openmeeg/issues/584 -- wrong inputs are not handled gracefully.

So how do we get the flexibility of Python (that people are rightly used to) with the speed advantages of C++? To me we should write all the "make sure the inputs are correct, and return/give people back what they expect" stuff in Python, and leave C++ for just the heavy lifting under the hood (keeping C++ classes private at the Python end as much as possible). Doing equivalent input validation in C++ would be a nightmare -- never as flexible, stable, maintainable, or *able to be fixed/tested/extended by others* as just doing it all in Python.

Hence to me the most maintainable thing is to essentially hide the C++ bits at the Python side, and write clean, simple, Python APIs from the start. (Well, probably from a 2.0 release since it's a big backward compat break that we hopefully only need to make once.) And we should start with the simplest case (MEG+EEG and 1- and 3-layer models?) and build up from there based on use cases.

> Maybe you can show me an example of what you mean.

In MNE-Python we have the "glue code" in place to leverage the C++ classes. I think we could start by moving this to OpenMEEG. In our public [`mne.make_forward_solution`](https://mne.tools/stable/generated/mne.make_forward_solution.html), at the end of the day it takes ndarrays of positions and mesh definitions (often stashed inside `dict`s) and gives you back the gain matrix in ndarray form (also stashed in a `dict` with metadata). You can see how we make use of the C++ classes here:

https://github.com/mne-tools/mne-python/blob/240df830bda18dc630f9bffea386ded3bd247ace/mne/forward/_compute_forward.py#L830-L874

Basically I propose that we add some more readable version of this to openmeeg that operates on ndarrays as much as possible.

We'd have to think about the API, but just to get the ball rolling:
```
def make_forward_solution(rr, meshes):
    """Compute a forward solution.

    Parameters
    ------------
    dipole_rr : shape (n_vertices, 3)
        The dipole positions.
    meshes : dict
        The meshes to use defining the conductive layers.
        Keys can be "brain", "skull", and "scalp".
        Values for each key must be a dict with keys:

        `'rr` : ndarray, shape (n_mesh_vertices,3)
            The mesh vertices (floats).
        `'tris'` : ndarray, shape (n_mesh_tris, 3)
            The mesh triangle definitions (ints).
    sensors : TBD
        We'd have to figure this out, probably via some other definition,
        since it's not-trivial.

    Returns
    -------
    gain : ndarray, shape (n_vertices, 3, n_sensors)
        The gain at each vertex in each direction (X, Y, Z).
```

agramfort: rather than exposing all the C++ classes I would say we need to have a documented and tested python API that uses these classes. It can be a thin wrapper that just does some validation. So I would also say that this list should be replaced by pure python functions. Once public classes are exposed, removing them from public API is some work (deprecation etc.) so let's do the work when we need it.

larsoner: >  Again. Long term I agree... I have added what is currently necessary to reproduce the apps. 

Okay, but in the short term even, why add them to `__init__.py` at all? By adding them to `__init__.py` you make them part of the public interface. You don't need to do this to create apps. For example if a `openmeeg whatever` app that you want to create lives in `openmeeg/commands/_whatever.py`, then in that `_whatever.py` you can do at the top `from ..openmeeg_cxx import GainEITInternalPot ` and the class stays in private namespaces only, rather than being imported into a public one -- but we can still use it internally the app (and the user never needs to know).

PR #609, opened 2023-04-22
MAINT: Build for Python 3.11

larsoner:
1. Bump cibuildwheel version so that 3.11 wheels are built
2. Bump to 2.6.0 so that we can cut a release

Okay to merge and release when green @agramfort @papadop ? Without doing this Python 3.11 wheels are not available, and 3.11 has been out since Oct 22.


agramfort (approved):

agramfort:
just note that you bump the version here so it's a new release?

larsoner:
Yeah to the the next patch level. It should allow new wheels to get uploaded properly to PyPI

PR #610, opened 2023-04-23
MAINT: Bump version

larsoner:
Bump the CMakeLists version back to 2.5.9999 and use `setuptools_scm` for the Python version from now on so we don't have to manually increment versions.


larsoner:
Does not appear to like the nested directory structure or usage with `cibuildwheel`, so I'll just bump manually

PR #644, opened 2024-03-25
MAINT: Simplify build process a bit

larsoner:
Looking at how sklearn builds for NumPy 2.0 I think this should be simpler


larsoner:
Actually I think explicit version pins here are okay

PR #658, opened 2024-05-20
MAINT: Switch to native builds on macOS

larsoner:
I expect this to require some iteration... we'll see!


larsoner:
Okay in the end I ended up:

1. Building wheels and apps on native macOS ARM
2. Testing wheels and apps on macOS ARM (because it's possible and easy now!)
3. Using `setuptools_scm` to set the `openmeeg.__version__`
4. As part of (3), I moved a bunch of stuff out of `setup.py` and into `pyproject.toml`. The only stuff left in `setup.py` has to do with building / SWIG'ing -- it's not obvious how to move that.

It was a lot of commits but the code ultimately is cleaner. Once it's green I'll adjust the merge requirements and merge!

PR #660, opened 2024-05-22
FIX: Hotfix for git archival

larsoner:
Unexpected behavior with git archival and setuptools_scm:
```
larsoner:~/Desktop/openmeeg-2.5.9/wrapping/python$ python -m setuptools_scm -c pyproject.toml
('ERROR: no version found for', Namespace(root=None, config='pyproject.toml', strip_dev=False, command=None))
```
so instead let's do
```
larsoner:~/Desktop/openmeeg-2.5.9$ python -m setuptools_scm -c wrapping/python/pyproject.toml
2.5.9
```

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -545,14 +545,16 @@ jobs:
           fi
           cd wrapping/python  # in root
           echo "path=wrapping/python/dist/*.whl" >> $GITHUB_OUTPUT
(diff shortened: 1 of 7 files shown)


larsoner:
Okay this also fixes a problem where we were inadvertently modifying LICENSE.txt when getting OpenBLAS using the NumPy scripts. Neat that `setuptools_scm` ended up finding a little bug for us by marking the source tree as dirty!

PR #708, opened 2024-12-11
MAINT: Switch to abi3 mode and unify before_all

### LINKED ISSUE #707: python 3.13 compatibility
I can get openmeeg 2.5.12 built against py3.13 from conda-forge, but not from PyPI, which is causing `uv` to choke:

```
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of openmeeg are available:
          openmeeg<=2.5.5
          openmeeg==2.5.6
          openmeeg==2.5.7
          openmeeg==2.5.8
          openmeeg==2.5.10
          openmeeg==2.5.11
          openmeeg==2.5.12
      and openmeeg>=2.5.5 has no wheels with a matching Python version tag, we can conclude that openmeeg>=2.5.5,<2.5.6 cannot be
      used.
      And because mne[full] depends on openmeeg>=2.5.5 and your workspace requires mne[full], we can conclude that your workspace's
      requirements are unsatisfiable.
```

This is not high-priority --- I'm experimenting with `uv`, not relying on it for daily work --- but it would be nice to have this work eventually.

larsoner:
Closes #707

Will cut a 2.5.13 immediately after this is merged


larsoner:
Okay received confirmation from the NumPy folks that abi3 should be supported, so I'm going to switch to that in this PR. It's a bigger change but should make maintenance easier

larsoner:
Now also adds a long-overdue unification / DRY of `cibw_before_all.sh` and `cibw_before_all_apps.sh`

PR #715, opened 2025-01-07
TST: Test pytest

larsoner:
Let's see if we can remotely debug https://github.com/pytest-dev/pytest/issues/11884#issuecomment-2574158153


larsoner:
Closing for now since I think using `--pyargs` is good enough for us. Will reopen if the upstream issue could use more testing this way

PR #798, opened 2026-01-10
Add FlexiBLAS implementation

papadop:
The fedora downstream seems to favor flexiblas for Blas/Lapack implementation. Add it natively.
In the process, I corrected a small inconsistency for  DGESDD.


larsoner (approved):
Worth trying to get a CI to install + test? There is at least https://formulae.brew.sh/formula/flexiblas so maybe it's not too hard to add a macOS run... but okay with me if you want to self-merge as is without it @papadop

PR #839, opened 2026-06-11
Build in SABI mode when using cmake

larsoner:
1. Add `PYTHON_USE_SABI` defaulting to ~~True~~ False to build in ABI3 mode when invoking `swig` from `cmake` (our wheels invoke `swig` from `setup.py` which already does this)
2. Avoid explicitly link to Python lib, allow `swig` to handle this (hopefully -- otherwise triage link based on ABI3 mode or not)
3. Switch from deprecated `swig_link_libraries` to `target_link_libraries`

Testing patch in parallel in https://github.com/conda-forge/openmeeg-feedstock/pull/78


larsoner:
https://github.com/conda-forge/openmeeg-feedstock/pull/78 was green so I'll add a test case here and mark for merge-when-green!

Issue #577, opened 2022-08-26
bug in openmeeg single layer MEG leadfield?


agramfort:
doing the integration of OpenMEEG in MNE and comparing solution between MNE linear collocation method and OpenMEEG results seem pretty different when using a 1 layer BEM model.

See https://gist.github.com/agramfort/c9dd2a643a5e02155c23aca915351e79

For a 3 layers model the MAGs and RDMs really good. But for 1 layer (MEG only) it's not great.

I have no intuition why. Although using a finer mesh for the single interface helps (see notebook at the bottom) it's still pretty bad...

@papadop @mclerc if you have any idea why this can happen let me know

also I think we do have any test in CTest/CMake except for 3 layers models. @papadop do you confirm?

papadop:
Indeed, there are no 1 layer models in the tests. It's not strictly only three layers models (as there are several non-nested models or models with zero conductivity domains).

papadop:
From the gist, it seems to be a realistic model. Can we go back to a spherical model and check against a ground truth ?
What makes you thinks it's OpenMEEG that is faulty (I'm just asking, there is effectively a good chance it is OpenMEEG as MNE is much more used, but I'm just wondering if there is a specific reason to suspect OpenMEEG).

agramfort:
See https://gist.github.com/agramfort/c9dd2a643a5e02155c23aca915351e79

OpenMEEG topo seems more noisy. 

Yes we should go back to a sphere model to validate MEG single layer bem 

Can you help?

papadop:
Do you have the script that you used to compute the 'fwd_om_meg.npy' and 'fwd_mne_meg.npy' models ?

agramfort:
I used this branch https://github.com/mne-tools/mne-python/pull/11011 but let's see if the bug fixes in Python wrappers magically fix the problem we are seeing

papadop:
Can we just generate the leadfields with the C++ api then just to rule in/out the fact that it is a core bug vs a python bug ?

larsoner:
> Can we just generate the leadfields with the C++ api then just to rule in/out the fact that it is a core bug vs a python bug ?

I just modified https://github.com/mne-tools/mne-python/pull/11011 to use a file-based interface instead of constructing the Geom in Python. It still fails in that case. I could try constructing the sensors and dipoles the same way if it would help.

papadop:
OK. I just generated the MEG leadfields for Head1, Head2 and Head3.
[...]

PR #851, opened 2026-07-16
TST: Sarvas test

larsoner:
@papadop up to you if you want this! It's what Claude Opus 4.8 came up with for a unit test.

Related to https://github.com/openmeeg/openmeeg/issues/577


papadop:
Thank you for this. I have not abandoned it... I have a branch with many, many modifications, which I need to split in smaller pieces. In particular, I found a better way to implement the ferguson operator (with the help of an LLM).

In the middle of that I corrected some awkward things that made the code not always coherent (e.g. we had two ways of computing a triangle normal with opposite directions). Hopefully, I will merge all of this in the coming weeks. 

But having the elements you provide is very helpful.... I have not looked at it in details, but adding the test seems a good idea to me. For the deflation, that might be a good idea too, especially if it always improves the accuracy. If it improves it for coarse meshes but make it worse for fine meshes, the change would be questionable. So i'd say, if it's not too much work,  let's try it and see what we get from it....

papadop:
The failures seem to have no relation with the proposed change...

larsoner:
Fixed the CI errors by "releasing" libaec in my own `vcpkg` fork to work around its persistent download errors. I'll mark this one for auto-merge because it's simple enough I think!

PR #852, opened 2026-07-16
ENH: Build with 64-bit OpenBLAS

larsoner:
This turned out to be pretty simple actually!

Closes #818

@papadop there are still things like `Dimension` and `Index` that are 32-bit, want me to change those to 64-bit while I'm at it?

Claude Opus 4.8 helped with the CMake transition and authored the test.


papadop:
I'm not very used to the wheel builds, so I may be wrong in my comments, but here they are.

- It looks to me that you totally disable the 32 bit build with this change. I would instead detect whether the OS is 32 or 64 bits and would adjust the library accordingly. That being said, I agree that 32 bits OS are more and more exceptions, but there are still some of them and it seems to me that with this patch, we no longer support them (https://itsfoss.community/t/summary-linux-distributions-for-32-bits/15398). Not sur how many users we have on such systems.
- I'm a bit annoyed by the fact that there are cmake variables such as USE_SCIPY_OPENBLAS that are not exposed to the cmake "user" interface. Wouldn't it be more logical to directly add a SCIPY_OPENBLAS mode and expose it through BLA_IMPLEMENTATION, so that people can try to use it directly at the cmake level ? We can do that later if you think this is a reasonable idea....
- For the Dimension an Index, I do not think that having matrices with indices bigger than 32 bits is meaningful. Could we have meshes that big ? That would increse memory for no purpose in my view. But, checking that the sizes of the matrices we create remains in that domain (i.e < 4294967296) might be a good idea. Again, there remains the question of whether we decide to drop the support of 32bit systems or not.

papadop:
And there are some build failures to understand...

larsoner:
> It looks to me that you totally disable the 32 bit build with this change

Should be only when using SciPy's openblas to build, which ...

> USE_SCIPY_OPENBLAS

... is really something that only the wheel build should be using (at least in my mind). So I'm not sure it needs to be user-exposed. Hence why I also wanted to just transition to 64-bit BLAS/LAPACK with no 32-bit option here

> I would instead detect whether the OS is 32 or 64 bits and would adjust the library accordingly

We could, but we don't currently build wheels for any 32-bit OS (and don't foresee adding it) so I don't see this being very useful / worth the effort. I don't think we actually test any 32-bit OS support at the moment.

larsoner:
FWIW all the failures are preexisting `vcpkg` download failures except `setuptools`, which I'll look into now

papadop:
OK. I buy your arguments. Hopefully, I will merge it tomorrow after having a closer look at the MR. Thank you.

larsoner:
Hah restarting the CI fixed it... so it's a flaky test (OpenBLAS non-determinism maybe?). I'll push a fix for the tolerance bump anyway to make the test more robust.

larsoner:
@papadop okay to merge this one?

PR #853, opened 2026-07-16
Fix dtype and contiguous bugs in Python wrappers

larsoner:
Closes #817
Helps with #584

Used Claude Opus 4.8 to help with this, but I understand (at least as much as I do with anything `swig`-related!) and vouch for the code.


larsoner:
This one seems pretty straightforward and tests are the real checkers here. CIs are green so in it goes!

PR #855, opened 2026-07-19
Minor changes

papadop:
Mainly some formatting changes:

- "for(" -> "for (".
-"if(" -> "if (".
- space changes.
- Some C++ 11 loops introduced.
- constness changes.
- Removal of some global temporaries.
- ...

No functional changes.


larsoner:
Looks like one test fails needing a tolerance bump

papadop:
Not really. I cannot reproduce the error on my computer. The errors I get are 10 times smaller. I do not see what change can lead to this. Investigating...

papadop:
On my machine the rdm/mag errors are:
```
[0.00611151 0.00399548 0.00328253 0.00285101]
[0.00182135 0.00221592 0.00290041 0.00380684]
```
So they are way below 0.02... 
I cannot explain why we have higher (a bit Heisenberg like) errors. When I added a print of the arrays, the error totally disappeared. Otherwise, it appears in a varying number of tests.... Very puzzling. Either, it is an infrastructure problem, or we have a random bug in the python implementation with _om_meg_gain, which is difficult to reproduce....

papadop:
OK. It seems it is a swig/pypy problem with initialization of default parameters. I still do not understand why it was not triggering before. My changes should not have made any difference.

PR #862, opened 2026-07-23
Reorganize Triangle class and exploit it.

papadop:
This re-organize a bit how triangles are handled in a more consistent fashion (normal and area are always computed). In addition, some integral reformulations have been made. 

More simplification are possible: We compute triangle points from barycentric coordinates to recompute barycentric coordinates from the point coordinates. This will maybe be in a further change....


papadop:
Honestly, I do not understand what happens on MacOS/arm64. When I bump up the tolerance on the null radial dipole MEG test, it is the RDM test on a single sphere that starts to fail. It does not seem to be related to threads, because it seems presents even if I forcing (I hope)  single thread computation. And increasing tolerances by more than 10 seems a lot.

larsoner:
I have a macOS arm64 machine, taking a look now...

larsoner:
Okay looks like it might be due to an old accidental `||`-to-`&&` flip in #395 that left single-shell head matrices un-deflated and singular. This leads to a garbage inverse and the leadfield is only whatever roundoff happens to survive, making it vary by platform, BLAS, thread count, non-deterministic accumulation errors, etc. Pushing a commit now -- see if the logic makes sense to you!

What exposed it (aided by Claude Opus 4.8):

Five consecutive runs, test_meg_sphere_vs_sarvas[1-conductivity0], RDM per dipole depth (tolerance 0.02):

rdm [0.003202 0.016143 0.003346 0.004216]
rdm [0.003202 0.002871 0.003346 0.009485]
rdm [0.003202 0.008708 0.028338 0.004706]   <- fails (0.0283)
rdm [0.003202 0.003493 0.003498 0.004037]
rdm [0.038081 0.003913 0.0076   0.006395]   <- fails (0.0381)

Two back-to-back ctest -R single_layer runs on the same main build disagree on both the values and which tests fail:

run 1:  x: array([0.003678, 0.062135, 0.003863, 0.004381])   ->  2 failed, 1 passed
run 2:  x: array([0.096637, 0.017719, 0.008718, 0.03479 ])   ->  1 failed, 2 passed

test_meg_sphere_radial_dipole_is_silent on main, three runs (tolerance 0.02, which is why the mac/arm workaround was raised to 0.15):

radial/tangential ratio = 0.00858
radial/tangential ratio = 0.08051
radial/tangential ratio = 0.01324

Pinning to one thread makes main reproducible — five identical runs:

OMP_NUM_THREADS=1
rdm [0.003202 0.002871 0.003346 0.004037]   (x5, bit-identical)

The threading randomizes which garbage we get. The head matrix is singular either way (M @ 1 = 1.5e-17 vs ‖M‖ = 0.058), so the inverse is meaningless on every platform, `main` just happens to reshuffle the roundoff that survives into the leadfield, while the branch's reformulated integrals pin it to one consistently-bad draw.

papadop:
Nice !  Thank you.... I really need to have an easy access to a MacOS platform.
Do you think the bug might be also related to issue #577 ?

larsoner:
I think with this fix in we can probably close that issue. The remaining error is (I think) attributable to the coarseness of the mesh. Maybe we should add a test with a finer mesh to confirm?

papadop:
Yeah. It is a good idea to add a test with a less coarse mesh. I'll merge this PR now. On my list are more things to do:

1. Correct CGAL stuff.
2. Eventually improve a bit more the integrator to have a mode where barycentric coordinates are passed to the function a nd avoid recomputing them.
3. Add a python interface (and maybe some C++ code which might be missing) to get blocl matrices and coefficients.
4. Use the previous interface to add a python routine that can efficiently approximate lead field matrices for different sets of conductivities.

larsoner:
Sounds good! Happy to help un-break macOS ARM stuff as needed 😄

PR #863, opened 2026-07-23
Bump OpenBLAS to version without DSYMM bug

larsoner:
There's some thread oversubscription that slows stuff down, but it seems like the most likely culprit here is linalg ill-conditioning in `hminv = om.HeadMat(geom, integrator).inverse()`. Going to see if this fixes things and keep digging a bit.


larsoner:
I've restarted the job a few times and haven't observed a failure. But let's leave this open for a bit in case it resurfaces

PR #866, opened 2026-07-26
Make the i/o layer thread-safe

larsoner:
I had Claude Opus 5 help me look into making Python 3.14t wheels, which requires looking into the OpenMEEG codebase for compatibility. It found some global state stuff in MathsIO, MeshIO, and GeometryIO. These are pretty much all fixable by passing through filenames rather than using a global state to track them. I think we can rely on CIs to ensure this stuff is okay. Claude wrote the changes and I reviewed them -- they look okay to me (but I'm not a C++ expert!).

@papadop feel free to review and merge if you're happy.

This is the more verbose description provided by Claude:

<details>
<summary>Changes</summary>

The maths and mesh i/o layers rely on shared mutable state that only the GIL was keeping safe. These are real races today for anyone using threading.Thread, and they are a prerequisite for ever declaring Py_MOD_GIL_NOT_USED for the free-threaded builds of CPython, since that declaration is a promise CPython takes at face value.

* MathsIO.C: ReadTag() read every file header into a function-level static buffer, so concurrent readers corrupted each other's format detection. Use a local buffer.

* MathsIO: the registered ios() are process-wide singletons, and operator>>, operator<< and info() each did io->setName(name) before dispatching, so the filename of a shared object was swapped for the duration of every operation. Two concurrent reads could therefore read the wrong file. Pass the filename to info()/read()/write() instead and drop setName().

* MathsIO: DefaultIO and permanent implement the one-shot format() override, and GetCurrentFormat() mutates DefaultIO to consume it. Make both thread_local: sharing them races, and one thread's format override should not be consumed by another thread's operation.

* MatlabIO: matio and the HDF5 library behind MAT 7.3 keep non-reentrant global state and are not built thread-safe here. Serialise info()/read()/write() on a private mutex. This also covers the shared static scratch buffers in details::helper<> (dims[] and saved), which are only reachable from those methods.

* MeshIO, GeometryIO: the base constructor inserted into the static registery, and create() reaches it through clone(), so every mesh or geometry load mutated a map that create() concurrently reads. Only the prototypes (the instances default-constructed with an empty filename) register now. The insert from a clone was already a no-op because the prototype owns the key, so behaviour is unchanged, and the registery is immutable once static initialisation is done -- concurrent reads then need no lock.

* logger.h: verbosity is written by set_info_level() while other threads read it, so make it atomic. It was also never initialised by the constructor and only worked because the singleton has static storage duration and was zero initialised first; initialise it to DEBUG explicitly, which is the value it previously ended up with. nullstream is written to by its callers, so make it thread_local.

The compute paths were audited for shared mutable state and are clean: what is left in OpenMEEG and OpenMEEGMaths is static member functions, thread_locals, the now-immutable registeries, and the state under the matio mutex.

Note that this changes the MathsIOBase virtuals and removes MathsIO::setName(). Both are on exported classes, but setName() had no callers outside MathsIO.C and the Python wrappers do not expose either.

Verified on macOS arm64: builds clean with -Werror -Wextra -pedantic and 692/692 ctest tests pass.

</details>

### DIFF
--- a/OpenMEEG/include/GeometryIO.h
+++ b/OpenMEEG/include/GeometryIO.h
@@ -71,12 +71,10 @@ namespace OpenMEEG {
 
         static Registery registery;
 
-        //  Only prototypes register, see MeshIO.h.
+        struct Prototype { };  //  Only the prototypes register, see MeshIO.h.
 
-        GeometryIO(const std::string& filename,const char* name): fname(filename) {
-            if (filename.empty())
-                registery.insert({ name, this });
-        }
+        GeometryIO(Prototype,const char* name) { registery.insert({ name, this }); }
+        GeometryIO(const std::string& filename): fname(filename) { }
 
         std::string fname;
     };
--- a/OpenMEEG/include/GeometryIO.h
+++ b/OpenMEEG/include/GeometryIO.h
@@ -71,7 +71,12 @@ namespace OpenMEEG {
 
         static Registery registery;
 
-        GeometryIO(const std::string& filename,const char* name): fname(filename) { registery.insert({ name, this }); }
+        //  Only prototypes register, see MeshIO.h.
+
+        GeometryIO(const std::string& filename,const char* name): fname(filename) {
+            if (filename.empty())
+                registery.insert({ name, this });
+        }
 
         std::string fname;
     };
--- a/OpenMEEG/include/GeometryIOs/GeomFile.h
+++ b/OpenMEEG/include/GeometryIOs/GeomFile.h
@@ -140,7 +140,9 @@ namespace OpenMEEG::GeometryIOs {
             return res;
         }
 
-        GeomFile(const std::string& filename=""): base(filename,"geom") { }
+        GeomFile(): base(Prototype{},"geom") { }
+
+        GeomFile(const std::string& filename): base(filename) { }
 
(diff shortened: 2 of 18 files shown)


papadop:
I just did a quick look. I'll have a more thorough one on Tuesday. From what I see, this is mostly nice suggestions, but I'm not very happy with passing the filename constantly. I'd prefer a specific constructor for prototypes (this is usually what I'm doing), but I need to better understand the Claude proposal before deciding. All the mutex and threadsafe proposal seem nice (as far as I can tell as I'm not very used to handle threads in C++ besides OpenMP). These IO interfaces were never meant to be used concurrently, and this is a unexpected outcome of the python interface, but a nice one.

larsoner:
> but I'm not very happy with passing the filename constantly. I'd prefer a specific constructor for prototypes (this is usually what I'm doing), but I need to better understand the Claude proposal before deciding.

I pushed a commit to try this... I think it's probably better. But feel free to look and see if you agree!

papadop:
I misread completely the patch about "passing the filename constantly". It is actually a nice simplification.
I do not quite like the Prototype stuff, but I will push a change in favor of changing the "prototype" construction to use a dedicated type that is explicitly named Extension (which is the meaning of the const char* it takes).

PR #868, opened 2026-07-27
Modernize macOS ARM support using Accelerate

larsoner:
In MNE-Python we have been doing some work to optimize linalg threading (https://github.com/mne-tools/mne-python/pull/14103, https://github.com/mne-tools/mne-installers/pull/460) and find that Accelerate on macOS arm64 outperforms OpenBLAS a lot. This is why NumPy and SciPy have switched to it for their linalg backend in their wheels. This PR adapts the NumPy+SciPy approach (using Claude 4.8 Opus to adapt / draft changes) and switches the wheel and app backends to use Accelerate.

Switching the wheel builds (which are tested) and adding a `thorough` build_and_test run should verify that things work properly here!

Will conflict with #852 but this one is a bigger deal so I'd rather land it first if possible. Would be good to cut a release shortly thereafter, maybe once the current batch of PRs is sorted out.


papadop:
I think the implementation would be much cleaner if instead of adding a FC_ALIAS(xxx); after each BLAS(xxx,XXX) and LAPACK(xxx,XXX); we changed the macros BLAS and LAPACK themselves to incorporate directly the content of the FC_ALIAS macros as:
```
#define BLAS(x,X) FC_GLOBAL(x,X)
#define LAPACK(x,X) FC_GLOBAL(x,X)
```
```
#define ACCELERATE_ASM_MARKER(x) __asm("_" #x "$NEWLAPACK")
#define BLAS(x,X) FC_GLOBAL(x,X)  ACCELERATE_ASM_MARKER(x)
#define LAPACK(x,X) FC_GLOBAL(x,X) ACCELERATE_ASM_MARKER(x)
#undef ACCELERATE_ASM_MARKER
```
in the file OpenMEEGMaths/include/BlasLapackImplementations/OpenMEEGMathsAccelerateConfig.h.
This will remove a lot of changes and the file OpenMEEGMaths/include/BlasLapackImplementations/FortranAlias.h would be no longer useful.

Am I missing something. Can you do that ? Otherwise, it can also be handled with a next PR, but I'd prefer if it was done with the current proposal....

larsoner:
Yes it was doable, feel free to look again!

papadop:
Better, but still there is some redundancy. We would need to remember adding a OM_NEWLAPACK(xxx) for every xxx lapack function that we would add. Why don't the change of the LAPACK and BLAS  (or directly FC_GLOBAL) macros like I proposed above ? The pragma solution seems better than the asm one, but maybe less portable (though it covers gcc/clang which is all we care for MacOs), but seems some over engineering.

To me, the simplest thing is to do:
#define FC_GLOBAL(x,X) x ## __##x##$NEWLAPACK
with an eventual OM_STR around "x ## __##x##$NEWLAPACK" and that's it.
No need to name the function one way (x ## _) and tell the compiler to rename it later with the pragma.

larsoner:
🤦 makes sense, pushed, saved about 100 lines (and of course the manual updating)!

papadop:
Yeah. Much simpler !!!
The random (or seemingly random) failure triggers in again. Looking at it with another branch/pr (not to be merged).

larsoner:
@papadop see also https://github.com/openmeeg/openmeeg/pull/863, it has some debugging magic in it ... I've restarted the job there as well. In the meantime I'll merge this so I can rebase / fix conflicts with #852

papadop:
Yeah. I just doubt that bumping the tolerances is the solution.... The reported errors are so much higher than the ones I see on my local machine that it must be algorithmic somehow. One advice I got from Copilot is to fix the number of threads (OpenBLAS, OMP, MKL, ...) to 1. It says that there are some computation inconsistencies with more than one thread (which I do not believe a lot). I do not see anything particularly special with the single layer test which fails regularly... And imposing one single thread will make the checks much much slower.

PR #872, opened 2026-07-29
DO NOT MERGE: Dump Sarvas intermediates from CI to localize flaky 3-layer RDM

larsoner:
I had Claude Fable 5 write up some remote instrumentation let's see if it tells us anything...


larsoner:
@papadop I went down a rabbit hole inspired by some bugs I hit ages ago on SVD decomp and found the problem:

---

Exactly the machinery you remembered: Intel SDE (sde64 -cpx is Cooper Lake) plus OPENBLAS_CORETYPE=Cooperlake, which 0.3.20 honours via force_coretype(). In the jammy container on my AVX512-less i7-7700K:

auto (Haswell):        max RDM = 0.006112  pass
Cooperlake (SDE -cpx): max RDM = 0.033539  FAIL

It's DSYMM, not the inverse

Dumping every array under both kernels localises it precisely. HeadMat, Head2MEGMat, DipSource2MEGMat, primary — bit-identical. ref, HeadMatInv, DipSourceMat, secondary — rounding only (1e-15). But gain differs structurally, 4.6% relative.

The tell: secondary = (Head2MEGMat @ Ainv) @ DipSourceMat computed in numpy is correct under both kernels, while OpenMEEG's GainMEG — same inputs, but through OpenBLAS — is wrong. So my earlier suspicion of DSPTRF/DSPTRI was misplaced; the inverse residual stays at 1.9e-16 throughout. Matrix::operator*(const SymMatrix&) at matrix.h:409 calls DSYMM(CblasRight, CblasUpper, …).

Minimal OpenMEEG-free reproducer (openmeeg-debug863/dsymm_bug.py, ctypes → cblas_dsymm):

0.3.20  kernel Haswell     max rel err 1.0e-15   ok
0.3.20  kernel Cooperlake  max rel err 1.4e+00   *** WRONG ***   (every shape tried)
0.3.32  kernel Cooperlake  max rel err 1.1e-15   ok

So: OpenBLAS 0.3.20's Cooper Lake DSYMM returns garbage, and it's fixed by 0.3.32. Your "0.3.20 bug since fixed" reasoning was right — it just runs through DSYMM rather than the SVD-style routines, and only bites on the subset of runners whose CPU dispatches to Cooperlake. That also explains cleanly why cibuildwheel (scipy-openblas), ubuntu-24.04 and the OpenBLAS-latest job never show it.

papadop:
Great. Thank you very much !!!

PR #873, opened 2026-07-29
Fix threading bug on MacOS.

papadop:
I wonder whether this is not the origin of the problems we sometimes see  on MacOS (the Ubuntu ones are different).
The comment before the ifdefs seems contrary to what has been implemented (I would have expected an ifdef). Furthermore, it is logical to have the `#pragma omp critical` inconditionnally, because those are OpenMP loops on triangles that update matrix elements indexed by vertices. So there is a potential thread race if two triangle with the same vertex are treated by two different threads. Chances that this triggers are low, but the fix seems logical and the recent test instabilities might stem from this.


larsoner:
If you merge `main` into this branch or rebase (one of them you'll need to do anyway to satisfy merge reqs now) the 22.04 Ubuntu error should go away

larsoner:
Seems reasonable to me. I had Claude look and it had an idea, feel free to completely ignore or read and maybe address (below).

Separately, currently no CIs use OpenMP on macOS, so I don't think we could hit this. But that's bad in and of itself. I have a macOS machine I can test with, want me to see if Claude and I can get it working again?

---

## Claude idea

git log -S shows the #ifndef __APPLE__ came from #582 (Sept 2022, the same PR that introduced ThreadException) — it wasn't arbitrary, it was papering over a real clang crash by trading it for a silent race.

The underlying spec violation is still there: om_assert throws (OMassert.H hardcodes IGNORE_NDEBUG, so asserts are live in release too), rhs(i) asserts, and it's now called inside a critical region inside ThreadException's try. OpenMP requires an exception thrown inside a critical region to be caught within that same region. That's exactly the StackOverflow issue the deleted comment linked.

atomic sidesteps it entirely and is cheaper than serializing on one global lock:
```
for (unsigned j=0; j<3; ++j) {
    double& r = rhs(triangle.vertex(j).index());   // assert stays outside the atomic
    #pragma omp atomic update
    r += v(j)*coeff;
}
```

papadop:
Adopted the Claude suggestion. I wonder whether we really need to keep the assert in this specific case in OMassert.H...

larsoner:
As usual, Windows finds something "wrong"
```
      D:\a\openmeeg\openmeeg\OpenMEEG\src\operators.cpp(58,33): error C7660: '#pragma omp atomic update': requires '-openmp:llvm' command line option(s) [D:\a\openmeeg\openmeeg\build\OpenMEEG\OpenMEEG.vcxproj]
```

papadop:
Yeah... I thought `#pragma omp atomic`  is pretty standard. I do not understand why it requires any new compilation flag (besides enabling openmp). This one was supposed to be a simple one... I'm a bit tired and the more complicated PR are not advancing.

larsoner:
I can take a look if you want, I had to fight with this type of stuff when getting the wheels to build

larsoner:
@papadop the Windows change didn't end up being too bad, and I also pushed a commit to actually enable OMP on macOS wheels (and Windows). Feel free to merge if you're happy! 🚀 

macOS OMP is pulled from anaconda because it's a convenient source. Could also use https://mac.r-project.org/openmp/ but it was slightly more complicated for the same oldest-OS support.

PR #878, opened 2026-08-05
Add 3.14t wheels and modernize python wrapping

### LINKED ISSUE #874: Replace removed Python 2 C API macros for SWIG 4.5.0 compatibility
Hi all.
I'm proposing this patch for fixing the compatibility of `openmeeg-2.5.16` with `swig-4.5.0`:

```
diff -up openmeeg-2.5.16/wrapping/python/openmeeg/_openmeeg.i.swig45 openmeeg-2.5.16/wrapping/python/openmeeg/_openmeeg.i
--- openmeeg-2.5.16/wrapping/python/openmeeg/_openmeeg.i.swig45	2026-07-28 10:15:42.082507949 +0200
+++ openmeeg-2.5.16/wrapping/python/openmeeg/_openmeeg.i	2026-07-28 10:15:42.087239833 +0200
@@ -378,7 +378,7 @@ namespace OpenMEEG {
     // C++ -> Python
 
     %typemap(out) unsigned& {
-        $result = PyInt_FromLong(*($1));
+        $result = PyLong_FromLong(*($1));
     }
 }
 
```

The patch comes from a pull request for `openmeeg` package on Fedora:
https://src.fedoraproject.org/rpms/openmeeg/pull-request/5#request_diff

larsoner:
1. In https://github.com/swig/swig/pull/3089#discussion_r1882976180 it was suggested that we fix our typemaps, this finally does it!
2. Get rid of the old Python 2 cruft
3. Modernizes our `setup.py` usage (bumping setuptools req in the process). Drafted and iterated with Claude Opus 5. The addition of the typemaps in principle allows us to put back `-fvirtual` (one of the three optimizations enabled by `-O`, and one of the two you can use in ABI3 mode), but in practice this only removes 8 of the 1000+ wrappers, and those become slower (so not really an optimization for us)
4. Add Python 3.14t free-threaded wheels (and ensure we don't require GIL)
5. Fix minor data-ownership bugs in `Vector::array()` and related functions that forced end users to use `.copy()` to prevent junk data

Closes #874

@papadop I think we should maybe cut a release after this WDYT? Also, this ended up being a bit longer than I anticipated as digging into the free-threaded stuff and adding tests I hit a speedbump with `Vector::array()`, which I fixed... and then that caused me (and Claude Opus 5) to look more into all the `ndarray` refs and fix/unify them. Happy to split it out into several PRs if that helps review!

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -281,7 +281,7 @@ jobs:
         else
           pip install --upgrade --only-binary numpy "numpy>=2"
         fi
-        pip install --upgrade --progress-bar off --only-binary="numpy" -q setuptools setuptools_scm wheel twine threadpoolctl "swig>=4.2" $EXTRA_PIP
+        pip install --upgrade --progress-bar off --only-binary="numpy" -q "setuptools>=70.1" setuptools_scm twine threadpoolctl "swig>=4.2" $EXTRA_PIP
 
 
     - name: MKL setup - mkl via conda
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -87,6 +87,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # 3.13 exercises the abi3 wheel on an interpreter it was not built for;
+        # 3.14t exercises the separate free-threaded wheel.
+        python-version: ['3.13', '3.14t']
+        os: [ubuntu-latest, windows-latest, macos-15-intel, macos-latest]
         include:
           - os: ubuntu-latest
             arch: x86_64
@@ -99,7 +103,7 @@ jobs:
     steps:
       - uses: actions/setup-python@v7.0.0
         with:
-          python-version: '3.13'
+          python-version: ${{ matrix.python-version }}
       - uses: actions/download-artifact@v8
         with:
           path: dist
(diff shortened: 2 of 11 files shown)


larsoner:
@papadop if you're happy here let me know and I'll adjust the merge requirements then merge

papadop:
Not sure to totally follow all th changes and especially what problem is fixed by point 5. Globally )I mostly looked at the C++ and swig stuff, you know all the python stuff much better Than I do.
One small comment.

- SWIG already defines the SWIG preprocessor symbol. We might very well replace the SWIGPYTHON by just SWIG....
- Before releasing, I'd like a bit more time (probably 2-3 weeks), so that I can finish a few things. Notably, the CGAL stuff is ready I think, and I'd like to fix the license issue and everyone is on vacation right now, so it is difficult to get some feedback on the subject. Do we need to rush to do the release ? Otherwise I can finish things after, but making another release so close might be a bit strange....

OK to merge this PR....

larsoner:
Okay I'll look at the symbol issue then merge. No rush to release, within the next month seems reasonable. Longer and we should probably just release multiple times. The release process now is just a few clicks on GitHub plus maybe one version in a file to change (and actually I might see if cmake has tag based version resolution like setuptools...)

larsoner:
> SWIG already defines the SWIG preprocessor symbol. We might very well replace the SWIGPYTHON by just SWIG....

Looks like SWIG is only defined during the parse, whereas during compilation SWIG is not defined but SWIGPYTHON is, so full replacement didn't work

papadop:
> > SWIG already defines the SWIG preprocessor symbol. We might very well replace the SWIGPYTHON by just SWIG....
> 
> Looks like SWIG is only defined during the parse, whereas during compilation SWIG is not defined but SWIGPYTHON is, so full replacement didn't work

The LLM advices are not always good !!!!
Thank you for the work.

PR #879, opened 2026-08-05
Fix memory leaks in Python interface

larsoner:
I found this one by debugging a bit... but `SWIG_PYTHON_SILENT_MEMLEAK` would have found it if we didn't have it enabled. So remove it, and fix the leak call sites.

This one is pretty straightforward so I'll mark for merge-when-green


papadop:
Thank you. I never realized about this SWIG_PYTHON_SILENT_MEMLEAK... Probably, I did not want to see it !!! Thank you again.

PR #881, opened 2026-08-10
Automated unified version parsing

larsoner:
@papadop one thing I've used for a few years in Python to simplify the release process is use SCM-based versioning. Basically, if you tag+release a version like 2.5.17, the `openmeeg.__version__` shows up correctly that way if you build against that commit (and if the package builder knows about the tag at that commit, which I've already ensured is the case using `.git_archival.txt` etc.). If you are one commit ahead, it'll parse as `2.5.18.dev1` for example. This PR adds the same support for `cmake` for the C++ version.

This way, when we want to release, there is no need for any more PR-to-update-version, then click buttons on GH, then PR-to-restore-dev-version. Now it boils down to just clicking buttons on GH (and adjusting release notes there if you want, etc.).

This uses custom-rolled code drafted by Claude Opus 5 to match setuptools-scm behavior. An alternative would be to use something like `rpavlik/cmake-modules`, but I'm inclined to try these ~90 lines and see if they make our lives easier.


papadop:
OK. Thank you. Can you just add in a README-dev.rst the actions that have to be done to create a new version. As far as I understand, it is just doing a git tag, but I just wonder what is the proper time to do it (from your description it seems that this has to be done just before a release). The above is just for this PR, but there are maybe some other actions.

larsoner:
> As far as I understand, it is just doing a git tag, but I just wonder what is the proper time to do it (from your description it seems that this has to be done just before a release). 

Yeah that's the idea -- and that adding that tag happens via cutting a release via "releases" (there is a button on the main page), doing "Draft a new release", then hitting the "generate release notes" and hitting the green button here:

Screenshot 2026-08-25 at 5 12 24 PM

It makes it super easy to cut releases. The app and wheel builder automatically attach the binaries and wheels, and then you finalize the publish (it sets the release back to draft mode when attaching). You can edit notes to your liking at that time or later etc. It's what I use in as many repos as possible nowadays. It's what I used in previous releases here as well, but I had to do a commit before and after to change (only) the C++ library version. Python already gets the version automagically from the tag.

larsoner:
Okay pushed a dev doc!

papadop:
Merged. Thank you very much !

PR #898, opened 2026-09-18
Add 315

larsoner:
1. Add 315 wheels to cibuildwheel config
2. Make change that will make cp315t work once SWIG supports it (swig/swig#3551)

@papadop once these land I plan to cut a 2.6.0 on Monday if there are no objections. I think the Accelerate improvements alone are enough to justify this rather than using a 2.5.17!


papadop:
As we exchanged before, can we wait till the end of the month. I have a former student that did some possible extensions and we work currently in integrating them properly into OpenMEEG. Honestly, I unsure that we will deliver on time, but let give us a chance.... Otherwise, we will leave that for a further release.

